PR #646, opened 2014-11-11
add fundraiser dependency to salesforce genmap

matt-thomas:
Line 20 of https://github.com/JacksonRiver/springboard_modules/blob/7.x-4.x/salesforce/salesforce_genmap/includes/modules/webform.inc calls a fundraiser hook but fundraiser isn't listed as a dependency of the salesforce_genmap module. This pull request makes it a dependency.


matt-thomas:
I've wrapped the hook in module_exists. Please let me know if I can do anything else here.

PR #672, opened 2014-12-30
Fundraiser sustainers simpletest improvements

dbarbar:
- Changed the way the Sustainers test calculates dates to match the way Springboard does it. This was causing false positives when skipping past the end of February and a date getting moved into March.
- Renamed the Sustainers and Date Mode test classes to match conventions.
- Moved those two test classes into the Fundraiser Sustainers test group.
- Made some test message more verbose for easier troubleshooting.
- Removed this inclusion of a class file that will get autoloaded instead.
- Commented out some empty test methods in the Date Mode test class so it won't waste time with setUp and tearDown on empty tests.

The date mode test has an "illegal string offset" error in PHP 5.4 where it's expecting $order to be an array, but instead it's a string of the order id. I couldn't easily and confidently track down what was intended, so I'll ticket that to address it later.


dbarbar:
@epicflux This is back up to date with 4.x as of July 14, 2015.

PR #673, opened 2015-01-07
Remove hardcoded refunded status so Partially refunded status will sync.

epicflux:
With this line removed a properlly mapped 'Partially Refunded' stage will map correctly.

### DIFF
--- a/salesforce/salesforce_donation/salesforce_donation.module
+++ b/salesforce/salesforce_donation/salesforce_donation.module
@@ -570,7 +570,7 @@ function salesforce_donation_salesforce_genmap_map_fields(stdClass $donation, En
         $fields['Transaction_Date_Time__c']  = gmdate('c', $refund->changed);
         $fields['CloseDate']  = gmdate('c', $refund->changed);
       }
-      $fields['StageName'] = 'Refunded';
+
       //$fields['Parent_Donation__c'] = '[Opportunity:donation:' . $donation->did . ']';
       $fields['Probability'] = 0;
       // Remove the donation id mapping if it's been set. No dupes for refunds!


salesforce/salesforce_donation/salesforce_donation.module
```diff
@@ -570,7 +570,7 @@ function salesforce_donation_salesforce_genmap_map_fields(stdClass $donation, En
         $fields['Transaction_Date_Time__c']  = gmdate('c', $refund->changed);
         $fields['CloseDate']  = gmdate('c', $refund->changed);
       }
-      $fields['StageName'] = 'Refunded';
```
pcave: This looks fine, but what happens if the 'partially refunded' status isn't mapped correctly. Just don't want to have opportunities end up with the wrong stage again.

epicflux: The partially refunded order status is mapped
in salesforce_donation_default_stage_map()

If its not mapped at all the default status is hard-coded to 'Posted'.
Which is done earlier in salesforce_donation_salesforce_genmap_map_fields()
at line 445.

We've added the alert to the status screen about unmapped stages.

We could change salesforce_donation_salesforce_genmap_map_fields() so
instead of forcing 'Posted' for all unset stages, we set the stage from the
default value in salesforce_donation_default_stage_map();



On Wed, Jan 7, 2015 at 12:09 PM, Phillip Cave <notifications@github.com>
wrote:

> In salesforce/salesforce_donation/salesforce_donation.module
> <https://github.com/JacksonRiver/springboard_modules/pull/673#discussion-diff-22599307>
> :
>
> > @@ -570,7 +570,7 @@ function salesforce_donation_salesforce_genmap_map_fields(stdClass $donation, En
> >          $fields['Transaction_Date_Time__c']  = gmdate('c', $refund->changed);
> >          $fields['CloseDate']  = gmdate('c', $refund->changed);
> >        }
> > -      $fields['StageName'] = 'Refunded';
>
> This looks fine, but what happens if the 'partially refunded' status isn't
> mapped correctly. Just don't want to have opportunities end up with the
> wrong stage again.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/JacksonRiver/springboard_modules/pull/673/files#r22599307>
> .
>



-- 
Jeremy Isett, Developer
Jackson River, LLC
www.jacksonriver.com
jeremy.isett@jacksonriver.com
mobile: 954.638.9479
home: 984.232.0048
skype: epicflux

PR #681, opened 2015-01-19
Simpletest fixes to setUp() method signatures that don't match their parent methods

dbarbar:
PHP 5.4 in strict reporting mode complains about these methods not matching their parent's signatures. You can see the errors on the Testing page after cleaning the caches.

Most of the Salesforce tests inherit from a class that declares setUp($modules, $permissions) and when they override that method they need to match its signature.

The same applies for the Fundraiser clone test and some of the webform user tests.

In most places I have added the extra arguments and passed them down into the parent so they'll work. This shouldn't affect the actual tests.

I'd almost argue that we don't need to pass modules and permissions up the class hierarchy like this, but that's something we can work on later.


epicflux:
Confirmed errors for the mis-configured tests no longer appear with these fixes.

PR #682, opened 2015-01-19
Fix and unblacklist the Email wrappers test case

dbarbar:
Fixed the email wrappers test case by adding a check for form state vars because they don't exist when doing the ajax template prefill.

Also discovered a bug with loading template for prefilling. Added a test for that and fixed it.

I took care to make the output for both fixes the same as the output before but now they won't cause errors and are more deliberate in setting NULL values when variables are undefined.


email_wrappers/email_wrappers.module
```diff
@@ -327,16 +327,21 @@ function email_wrappers_form_webform_email_edit_form_alter(&$form, &$form_state,
...
-  $keys = array('subject', 'from_address', 'from_name',);
+  $keys = array('subject', 'from_address', 'from_name');
   foreach ($keys as $key) {
     $fields = array('option', 'custom', 'component');
     foreach ($fields as $field) {
-      $form_state['values'][$key . '_' . $field] = $form_state['input'][$key . '_' . $field];
+      $full_key = $key . '_' . $field;
+      // Not all of these input items are available when using the AJAX
+      // prefill template selector.
+      if (isset($form_state['input'][$full_key])) {
```
dbarbar: This is the actual change. All other changes are whitespace.

On the email wrappers test, I was getting undefined notices for $form_state['input']['subject_component'] and $form_state['input']['from_name_component'] when it was doing a Post Ajax call to pick the template to use to prefill the email wrappers settings. I think that was triggering this validate function, and only the from_address_component was defined at the time.

email_wrappers/email_wrappers.module
```diff
@@ -851,8 +860,15 @@ function email_wrappers_load_defaults_from_template($nid) {
 
   // BCC gets special handling as it's the only field that may have
   // multiple values.
-  foreach ($bcc_email as $bcc) {
-    $bcc_addresses[] = $bcc['value'];
+  if (is_array($bcc_email)) {
```
dbarbar: When loading the bcc_email field, it could be either NULL or an array. When it was NULL the foreach would complain, and $bcc_addresses would never get initialized, causing an error with that. But the output would be correct because bcc_email would end up as NULL. So the output here is the same as before, but without the PHP errors.

email_wrappers/tests/email_wrappers_load_template.test
```diff
@@ -0,0 +1,67 @@
+<?php
```
dbarbar: This is a test for load template errors. It actually passes both before and after the fix, but before the fix it throws 2 exceptions, which I explained around the fix itself.

springboard/modules/springboard_tests/tests_blacklist.csv
```diff
@@ -5,7 +5,6 @@ DateMigrateExampleUnitTest
 CoderReviewStyleTest
 CoderUpgradeWebTestCase
 DsrTestCase
-EmailWrappersTestCase
```
dbarbar: Removing this test class name here means our automated testing will now run it.

epicflux:
Confirmed these tests pass.

PR #683, opened 2015-01-21
Change some market source p2p personal campaign fields to textareas

dbarbar:
Changed p2p ms referrer and p2p ms initial referrer to textareas in an update hook and updated the features.

This reverts the p2p personal campaign feature for fields when the update hook is run.


epicflux:
Re-opened in https://github.com/JacksonRiver/springboard_modules/pull/686

PR #684, opened 2015-01-27
Updated fundraiser commerce unit tests so they pass under drupal commerce

dbarbar:
- Updated the country and zone counts to match what drupal currently has.
 - Changed some ubercart specific order updating to match Drupal Commerce instead.
 - Cleaned up some coding style.


epicflux:
Confirmed the test now passes.

PR #685, opened 2015-01-29
Fix for Dynamic Gift String Default Value Handling

epicflux:
This issue:
If a gift string does not include the default value, and the field's default value is not in the passed amounts, a illegal choice error will appear on form submit if no amount is selected.

This fix:
For a dynamic gift string only set the default amount if it is passed and available in the options, else unset it.


dbarbar:
This worked for me. I ran the simpletests and tested manually using the token generate page. I did some cosmetic tweaks to the tests but didn't change the behavior.

PR #689, opened 2015-02-06
Throw Exception in Springboard Update Hook for missing Field Group module 

epicflux:
Throw exception in springboard_update_7001() if the fieldgroup module is not installed.


dbarbar:
This is perfect. The upgrade message is clear that it needs to be run again, and future updates are skipped.

I tested by temporarily removing the dependency on field_group, disabling field_group, and manually changing the schema update number in the system table. When I ran updb the exception was thrown and the update did not continue. I ran updb a few more times to confirm the same behavior. Then I enabled field_group and ran updb to confirm it was resolved. Finally, I ran updb again to confirm that no more updates were needed.

Works for me.

dbarbar:
BTW did you see entityreference_update_7000()? It attempts the module_enable inside the update. Is there a reason why we wouldn't just do that?

PR #698, opened 2015-03-02
Remove Springboard Feed during springboard_views uninstall, and edit it during install if it exists.

dbarbar:
Added an uninstall hook for springboard_views to remove the Springboard Feed, and added a check when adding the Springboard Feed to edit it if it exists.

I tested this change by disabling, uninstalling, reenabling springboard_views, and running cron via drush (at least 3 rounds of this) and checking the aggregator tables in between steps.  Uninstalling springboard_views will remove the feed and the feed's items.

### DIFF
--- a/springboard/modules/springboard_views/springboard_views.install
+++ b/springboard/modules/springboard_views/springboard_views.install
@@ -12,6 +12,23 @@ function springboard_views_install() {
 }
 
 /**
+ * Implements hook_uninstall().
+ *
+ * Removes the Springboard Feed.
+ */
+function springboard_views_uninstall() {
+  // Find the feed ID.
+  $fid = db_query("SELECT fid FROM {aggregator_feed} WHERE title = 'Springboard Feed';")
+    ->fetchField();
+
+  // Passing in only the feed ID will delete it and related items
+  // and categories.
+  if (is_numeric($fid)) {
+    aggregator_save_feed(array('fid' => $fid));
+  }
+}
+
+/**
  * Implements hook_requirements().
  *
  * Checks Views for anonymous access.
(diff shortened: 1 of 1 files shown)


springboard/modules/springboard_views/springboard_views.install
```diff
@@ -12,6 +12,23 @@ function springboard_views_install() {
 }
 
 /**
+ * Implements hook_uninstall().
+ *
+ * Removes the Springboard Feed.
+ */
+function springboard_views_uninstall() {
+  // Find the feed ID.
+  $fid = db_query("SELECT fid FROM {aggregator_feed} WHERE title = 'Springboard Feed';")
```
dbarbar: I confirmed that $fid will be false here if there is no matching feed. So the below logic won't complain.

springboard/modules/springboard_views/springboard_views.install
```diff
@@ -73,24 +90,34 @@ function springboard_views_requirements($phase) {
...
+ * Enables aggregator and views_data_export, and creates the Springboard Feed.
  */
- function springboard_views_update_7000() {
-   module_enable(array('aggregator', 'views_data_export'));
-   springboard_views_create_aggregator_feed();
- }
+function springboard_views_update_7000() {
+  module_enable(array('aggregator', 'views_data_export'));
+  springboard_views_create_aggregator_feed();
+}
```
dbarbar: Indentation fixes here and below.

springboard/modules/springboard_views/springboard_views.install
```diff
@@ -12,6 +12,23 @@ function springboard_views_install() {
...
+ */
+function springboard_views_uninstall() {
+  // Find the feed ID.
+  $fid = db_query("SELECT fid FROM {aggregator_feed} WHERE title = 'Springboard Feed';")
+    ->fetchField();
+
+  // Passing in only the feed ID will delete it and related items
+  // and categories.
+  if (is_numeric($fid)) {
+    aggregator_save_feed(array('fid' => $fid));
```
dbarbar: This function handles all aggregator feed add/update/delete operations, depending on what you pass in.

PR #700, opened 2015-03-09
Use configured submit button text when submitting through the API.

epicflux:
Issue:
The API forced the text 'Donate' when submitting donation forms. If the submit button isn't configured to use that word the submission fails to save.

Fix:
Follow the same practice found in other code, and webform itself, that uses the value set in the $node->webform['submit_text'] setting to determine the submit button's value.

Also made some minor changes to the tests to get that working.


epicflux:
@dbarbar please confirm the test runs and completes, I'll send this over to @kljr for confirmation the change to the API function looks good.

PR #704, opened 2015-03-18
Fixed error in which salesforce_genmap returned a fatal error.

nathaneanes:
As I noted briefly in the commit message, the salesforce_genmap returned a fatal error whenever I loaded the node/%/salesforce menu item on any webform node. This happens if the Fundraiser module is not installed. This is because it calls the function 'fundraiser_is_donation_type()', but the module does not have a hard dependency on fundraiser.module in its .info file. I added a module_exists() check to remedy this (rather than adding the dependency).

I discovered this in an installation in which I was using salesforce_genmap to map webform nodes to Salesforce, but without the need for the whole Fundraiser suite. I think this module_exists() check will remedy it, as it'll still let the $fundraiser variable be set to FALSE.


epicflux:
This was fixed in PR #646

PR #705, opened 2015-03-18
P2P progress discrepancy

dbarbar:
Added an extra check to the p2p donation success hook implementation to prevent it from running more than once on the same donation.


epicflux:
The code resolves the issue of Paypal transactions causing the p2p campaign total amount to increment incorrectly.

PR #711, opened 2015-03-31
Date Mode Setting Wrong Date when Transitioning Donations

epicflux:
When enabling/disabling the Date mode feature a donation series occuring on the last day of the month would have sporadic charge dates.

The function validating each donation day was not using the next_charge value to determine the day of the charge.

This fix pulls the next_charge value from the database when evaulating the days.

### DIFF
--- a/fundraiser/modules/fundraiser_sustainers/modules/fundraiser_date_mode/fundraiser_date_mode.module
+++ b/fundraiser/modules/fundraiser_sustainers/modules/fundraiser_date_mode/fundraiser_date_mode.module
@@ -188,12 +188,11 @@ function fundraiser_date_mode_max_records() {
  *
  * @param array $orders
  *   An array of sustainer series order records.
- *
- * @param int $new_date
+ * @param int|null $new_date
  *   Day of the month sustainers should be processed.
  *
- * @return array
- *   Returns an array of sustainer series records that have next_charge
+ * @return int[]
+ *   Returns an array of sustainer series Donation IDs that have next_charge
  *   values that will need to be updated based on the day selection.
  */
 function _fundraiser_date_mode_orders_to_change($orders, $new_date = NULL) {
@@ -225,6 +224,7 @@ function _fundraiser_date_mode_orders_to_change($orders, $new_date = NULL) {
 
     $selected_orders[] = $order['did'];
   }
+
   return $selected_orders;
 }
 
@@ -254,8 +254,10 @@ function _fundraiser_date_mode_calculate_delta($charge_time, $date) {
 /**
  * Verify all series orders fall on the same day of the month.
  *
- * @param array $series_orders
- *   Array of sustainer series order records.
+ * @param int[] $series_orders
+ *   Array of sustainer series Donation IDs.
+ * @param int|null $master_order_day
+ *   The day number of the month.
  *
  * @return bool
  *   Returns TRUE if all orders fall on the same day of the month.
@@ -285,6 +287,7 @@ function _fundraiser_date_mode_validate_order_dates($series_orders, $master_orde
       $dates[] = $date['day'];
     }
   }
+
   return count($dates) === 1 ? TRUE : FALSE;
 }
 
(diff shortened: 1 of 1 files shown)


fundraiser/modules/fundraiser_sustainers/modules/fundraiser_date_mode/fundraiser_date_mode.module
```diff
@@ -254,16 +252,31 @@ function _fundraiser_date_mode_calculate_delta($charge_time, $date) {
...
+ * @param int|null $master_order_day
+ *   The day number of the month.
  *
  * @return bool
  *   Returns TRUE if all orders fall on the same day of the month.
  */
 function _fundraiser_date_mode_validate_order_dates($series_orders, $master_order_day = NULL) {
   $dates = array();
-  foreach ($series_orders as $order) {
-    $date = _fundraiser_date_mode_explode_date($order['next_charge']);
```
dbarbar: This is the bug. This function originally thought $series_orders is an array of sustainer record arrays, but it's actually only an array of donation IDs. So it would attempt to extract a date from $order['next_charge'], which is NULL, and it would determine the date to be '1969/12/31/19/00/00'. Then it would determine that all sustainers were on the 31st and returns TRUE. If all sustainers actually are on the same day (day of month < 29) then the behavior is unintentionally correct. For the day of month > 28 this incorrectly returns TRUE and applies the same time delta to every sustainer.

The solution @epicflux implemented is to lookup the next_charge dates here and use them explicitly. An alternative option would be to pass the full sustainer records into this function as it was originally intended. That would avoid the extra db queries as we already have the next_charge data earlier in the process. On the other hand, this only happens in a drupal batch when enabling date mode or changing the date mode date, so performance isn't incredibly important here.

fundraiser/modules/fundraiser_sustainers/modules/fundraiser_date_mode/fundraiser_date_mode.module
```diff
@@ -515,18 +536,19 @@ function fundraiser_date_mode_update_date_mode_finished() {
...
+ * @param array $master_order
  *   Associative array containing information about the series master order.
- * @param mixed $series_orders
+ * @param array $series_orders
  *   Associative array of series orders.
  */
 function _fundraiser_date_mode_alter_next_charge($date, $master_order, $series_orders) {
-  $orders_to_process = _fundraiser_date_mode_orders_to_change($series_orders, $date);
+  // This is an array of Donation IDs.
+  $dids_to_process = _fundraiser_date_mode_orders_to_change($series_orders, $date);
```
dbarbar: I renamed this variable, only in this function, and added a comment to be very clear what this variable contains.

fundraiser/modules/fundraiser_sustainers/modules/fundraiser_date_mode/fundraiser_date_mode.module
```diff
@@ -254,16 +252,31 @@ function _fundraiser_date_mode_calculate_delta($charge_time, $date) {
...
+    ->fields('s', array('did', 'next_charge'))
+    ->condition('did', $series_orders)
+    ->execute()
+    ->fetchAllKeyed();
+
+  foreach ($series_orders as $did) {
+    $date = _fundraiser_date_mode_explode_date($next_charges[$did]);
+
     // If supplied, validate order days against the master order day.
     if ($master_order_day && $master_order_day != $date['day']) {
```
dbarbar: Before the fix, $date['day'] would always evaluate to 31.

fundraiser/modules/fundraiser_sustainers/modules/fundraiser_date_mode/fundraiser_date_mode.module
```diff
@@ -536,7 +558,7 @@ function _fundraiser_date_mode_alter_next_charge($date, $master_order, $series_o
     if ($orders_on_same_day) {
```
dbarbar: When all of the sustainers are on the same day (incorrectly evaluating to TRUE in some cases before the fix), the same time delta would be applied in bulk to all sustainers.

When they're not all on the same day, the time delta would be calculated and applied per sustainer.

IMO this is a bit of a premature optimization for a process that rarely happens, but we probably don't want to bother refactoring at this point.

dbarbar:
I added some clarifications and notes. This looks good to me. There are some improvements that could be made that I've noted, but none of it should hold up this fix.

PR #712, opened 2015-04-06
Update sb_social.addthis.js  ajax call to use Drupal.settings.basePath 

jesswalls:
If Drupal is installed in a subdirectory, the current ajax call to the 'sb_social' menu callback will produce a 404. This change prepends Drupal.settings.basePath to the ajax call to determine the correct subpath.


epicflux:
Tested and confirmed the Ajax request URI properly includes the base path.

PR #714, opened 2015-04-10
Donation form selectable currency support

dbarbar:
This adds Drupal Commerce currency support to the Fundraiser module. For the most part this is cosmetic only and hasn't been tested with the gateways.

The donor-side currency selection functionality has been removed from the multi currency module, and the ability for an admin to select the currency for a donation form has been moved into the Fundraiser module. The multi currency module is now a stub and Fundraiser includes an update hook to disable it. We can fully remove the multi currency module in a later version.

There is now a default site-wide currency, and the possibility to enable multiple currencies that can be selected when configuring the donation form. You can get to the settings by going to Gear -> Settings -> Manage Currencies in the Springboard Admin UI.

A new entity field for currency is added to fundraiser enabled content types. This is the dropdown populated by enabled currencies and defaulted to the site-wide default currency. This then creates and handles a currency webform field so currency is added to the webform submission data. It's also used to replace the other amount prefix, the quantity total (in JS) and ask amount labels (str_replace on $).

The webform component value gets updated when the currency node field is updated.

A currency JS setting is now available, plus new confirmation tokens. I updated the default confirmations to include the symbol token. I also replaced a lot of uses of 'USD' with commerce_default_currency() so in the future a non-USD currency could be used as the default.


fundraiser/fundraiser.info
```diff
@@ -4,7 +4,12 @@ core = 7.x
 version = 7.x-4.x-dev
 package = Jackson River Springboard
 configure = admin/config/system/fundraiser
+
 dependencies[] = token
+dependencies[] = commerce
+dependencies[] = list
+dependencies[] = entity
```
dbarbar: Directly depending on Commerce for currency stuff. List and entity are used for the currency field.

fundraiser/fundraiser.install
```diff
@@ -458,3 +459,53 @@ function fundraiser_update_7005() {
...
+function fundraiser_update_7008() {
+  if (module_exists('fundraiser_multi_currency')) {
+    module_disable(array('fundraiser_multi_currency'), FALSE);
+    drupal_uninstall_modules(array('fundraiser_multi_currency'), FALSE);
+  }
+
+  if (db_table_exists('fundraiser_currencies')) {
+    db_drop_table('fundraiser_currencies');
+  }
+}
```
dbarbar: I'm not opposed to putting these updates in one hook or handling the upgrade path differently.

fundraiser/fundraiser.module
```diff
@@ -152,6 +152,21 @@ function fundraiser_token_info() {
...
+  );
+  $info['tokens']['donation']['currency:name'] = array(
+    'name' => t('Currency name'),
+    'description' => t('The full currency name. For example, "United States Dollar".'),
+  );
+  $info['tokens']['donation']['currency:symbol'] = array(
+    'name' => t('Currency symbol'),
+    'description' => t('The currency symbol. For example, "$".'),
+  );
+
```
dbarbar: The main currency tokens.

fundraiser/fundraiser.module
```diff
@@ -165,7 +180,7 @@ function _fundraiser_donation_token_info($fields = NULL, $tokens = array()) {
   $children = element_children($fields);
   foreach ($children as $child) {
     // TODO Adjust this tokening to account for the new changed in donation fields.
-    $disallowed = array('card_cvv');
+    $disallowed = array('card_cvv', 'currency');
```
dbarbar: This skips adding the currency webform component as a token. We're handling that on our own.

fundraiser/fundraiser.module
```diff
@@ -211,6 +226,12 @@ function fundraiser_tokens($type, $tokens, $data = array(), $options = array())
         $donation->donation['payment_fields']['credit']['card_type'] : '';
     }
 
+    if (is_array($data['currency'])) {
+      foreach ($data['currency'] as $key => $value) {
+        $data['currency:' . $key] = $value;
```
dbarbar: Adds all the tokens available from the Commerce currency array.

fundraiser/fundraiser.module
```diff
@@ -1009,6 +1030,17 @@ function fundraiser_fundraiser_field_info() {
...
+
+  $fields['donation']['currency'] = array(
+    '#title' => 'Currency',
+    '#type' => 'hidden',
+    '#value' => '',
+    '#required' => 1,
+    '#allow_update' => 1,
+    '#reset_value' => 1,
+    '#create_callback' => 'fundraiser_currency_field_create',
+  );
```
dbarbar: This is the fundraiser/webform field definition.

fundraiser/fundraiser.module
```diff
@@ -1140,10 +1172,13 @@ function fundraiser_fundraiser_field_info() {
  * Field callback, called at runtime to update the field with fundraiser specific settings during creation.
  */
 function _fundraiser_amount_field_create($fundraiser, $field) {
+  $currency = fundraiser_get_currency_from_node($fundraiser);
```
dbarbar: Gets the currency array from either the node field, or the site-wide default.

fundraiser/fundraiser.module
```diff
@@ -1515,11 +1574,37 @@ function fundraiser_donation_form(&$form, &$form_state) {
...
+      if (isset($amount['#options'])) {
+        foreach ($amount['#options'] as $val => $label) {
+          $amount['#options'][$val] = str_replace('$', $symbol, $label);
+        }
+      }
+      else {
+        $amount['#field_prefix'] = $symbol;
+      }
+    }
+    $form = _fundraiser_update_form_field($form, 'amount', $amount);
```
dbarbar: We can remove this changing of the ask amount labels if we want to leave it to the admin.

fundraiser/fundraiser.module
```diff
@@ -3237,3 +3323,86 @@ function _fundraiser_gateway_status_alert($nodes, $method_type, $gateway_id, $st
...
+  return $currency;
+}
+
+/**
+ * Retrieves and formats the currencies for a form select element.
+ *
+ * @return array
+ *   The options array for a select form element.
+ */
+function fundraiser_currency_field_get_currency_options() {
```
dbarbar: This is a callback to populate the node currency field.

fundraiser/modules/fundraiser_commerce/fundraiser_commerce.module
```diff
@@ -242,28 +244,36 @@ function commerce_donation_line_item_add_form($element, &$form_state) {
...
+  // Ideally it should be the currency of the other
+  // line items or original donation.
+  $form['currency'] = array(
+    '#type' => 'select',
+    '#title' => t('Currency'),
+    '#description' => t('Enter the currency of the donation to add to the order.'),
+    '#options' => $available_currency,
+    '#default_value' => commerce_default_currency(),
+  );
+
```
dbarbar: This all is for adding a line item to an order. It's possible to add multiple line items of differeing currencies but the order gets weird. Not recommended. Plus the message below warns to not do this anyway.

fundraiser/modules/fundraiser_commerce/fundraiser_commerce.module
```diff
@@ -852,6 +862,7 @@ function fundraiser_commerce_fundraiser_donation_get_donation($donation) {
   // So we have the correct decimaled value.
   $donation->donation['amount'] = commerce_currency_amount_to_decimal($order->commerce_order_total[LANGUAGE_NONE][0]['amount'],
     $order->commerce_order_total[LANGUAGE_NONE][0]['currency_code']);
+  $donation->donation['currency'] = commerce_currency_load($order->commerce_order_total[LANGUAGE_NONE][0]['currency_code']);
```
dbarbar: This makes the entire currency array available to the donation.

fundraiser/modules/fundraiser_commerce/fundraiser_commerce.module
```diff
@@ -957,7 +969,7 @@ function fundraiser_commerce_fundraiser_donation_create($donation) {
   if ($donation->donation['amount'] == 'other') {
     $donation->donation['amount'] = preg_replace('/[^\d\.]/i', '', $donation->donation['other_amount']);
   }
-  $donation->donation['currency'] = isset($donation->donation['currency']) ? $donation->donation['currency'] : 'USD';
+  $donation->donation['currency'] = isset($donation->donation['currency']) ? $donation->donation['currency'] : commerce_default_currency();
```
dbarbar: Might need to double check this. We us currency code when saving (create/update) but the full currency array when getting. That's confusing.

fundraiser/modules/fundraiser_multi_currency/fundraiser_multi_currency.info
```diff
@@ -2,5 +2,8 @@ name = Fundraiser Multiple Currency
 description = "Adds multiple currency support to donation forms."
 core = 7.x
 package = Jackson River Springboard
-dependencies[] = fundraiser
 version = 7.x-4.x-dev
+
+dependencies[] = fundraiser
+
+hidden = TRUE
```
dbarbar: Leaving in the info file in case somebody somewhere has this module enabled. The fundraiser module will disable and uninstall it.

fundraiser/modules/fundraiser_multi_currency/fundraiser_multi_currency.install
```diff
@@ -1,7 +1,11 @@
 <?php
 /**
  * @file
- * Install, update and uninstall functions for the fundraiser_multi_currency module.
+ * The schema for the multi currency module.
+ *
+ * This module is deprecated. The schema is here so fundraiser can uninstall it.
+ * After it's been in this state for long enough, this module will be removed
+ * entirely.
```
dbarbar: Leaving this file since it has the schema definition which is needed for the uninstall.

fundraiser/modules/fundraiser_webform/fundraiser_webform.module
```diff
@@ -430,22 +430,21 @@ function fundraiser_webform_form_node_form_alter(&$form, &$form_state, $form_id)
...
+ * This is done with hook_form_alter and not hook_form_FORM_ID_alter
+ * because webform's form id varies.
+ * Ex. webform_client_form_NID.
  */
 function fundraiser_webform_form_alter(&$form, &$form_state, $form_id) {
-  // Alter the webform client form.
-  // Why hook_form_alter and not hook_form_FORM_ID_alter?
-  // Because webform's form id varies: webform_client_form_NID.
-  if (strstr($form_id, 'webform_client_form') !== FALSE) {
+  if (strstr($form_id, 'webform_client_form') !== FALSE && fundraiser_is_donation_type($form['#node']->type)) {
```
dbarbar: I reworked the logic here because it was testing the same conditions multiple times. Please double check that I got it right.

PR #740, opened 2015-05-20
Update the currency symbol on webform submission view and edit pages.

dbarbar:


epicflux:
@dbarbar this looked good.

I changed where the values are pulled from, instead of using the form values I set it to pull from the donation object where we now have ready to use values like 'formatted_amount'.

Let me know if my changes are okay.

PR #743, opened 2015-05-26
363 currency with fundraiser tickets

dbarbar:
- Uses the currency from the ticketed event node to display currency in the ticket boxes.
 - Forces the same currency when created ticket types.
 - Validates that the add-on donation form has the same currency as the ticketed event node.


fundraiser/modules/fundraiser_tickets/components/tickets.inc
```diff
@@ -181,6 +181,14 @@ function theme_fundraiser_tickets_webform_table($variables) {
 
   $element = $variables['element'];
 
+  $node = node_load($element['#webform_component']['nid']);
```
pcave: You sure this is always going to have a valid nid?

epicflux: This is a theme function for a webform component and the nid is one of the core values of a webform component. You can't have a component without that value.

PR #751, opened 2015-06-08
386 re enabling message action loses taxonomy

kljr:
Both the advocacy and message_action install files improperly referenced the taxonomy field names, omitting the "field_" prefix when checking field existence with field_info_field and field_info_instance. This caused a problem when message action module was disabled/re-enabled.

Also, the creation of the message action field instances was moved to the message action install file.

Hitching a ride in this patch is a syntax fix in two spots in which the number of existing messages was counted in order to determine whether the user-editable option should be disabled on the message add/edit page, and whether a message should be shown on the message landing page. The bad syntax did not result a bug - everything worked as expected - but it was obviously wrong.


kljr:
opened against wrong branch

PR #757, opened 2015-06-09
Create Sustainer Series on Cron

epicflux:
This adds a queue worker for creating the sustainer series on cron.

Also, there's an add admin setting for control over the cron options so an admin could choose to continue creating the series on donation submit.

### DIFF
--- a/fundraiser/modules/fundraiser_sustainers/fundraiser_sustainers.module
+++ b/fundraiser/modules/fundraiser_sustainers/fundraiser_sustainers.module
@@ -1327,12 +1327,18 @@ function fundraiser_sustainers_edit_form($did) {
     return t('You do not have access to this recurring donation.');
   }
 
+  // Text used for various sections of the form if no more charges remain.
+  $no_remaining_message = t('There are no further charges for this recurring donation.');
+
   // If the series creation is pending let the user know.
   if (
     !empty($donation->data['fundraiser_sustainers_series_status'])
     && $donation->data['fundraiser_sustainers_series_status'] == FUNDRAISER_SUSTAINERS_SERIES_STATUS_PENDING
   ) {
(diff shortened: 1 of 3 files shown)


fundraiser/modules/fundraiser_sustainers/fundraiser_sustainers.module
```diff
@@ -1204,11 +1227,20 @@ function fundraiser_sustainers_form_fundraiser_admin_settings_alter(&$form, &$fo
...
-      'standard cron run. These tasks will need to be cronned separately via sitename/fundraiser_cron'),
-    '#default_value' => variable_get('fundraiser_standalone_cron_enabled', 0),
+    '#title' => t('Sustainer processing on standalone cron.'),
+    '#description' => t('If this option is enabled the payment processing for recurring donations will removed from the ' .
+      'standard cron run. This task will need to be cronned separately via sitename/fundraiser_cron'),
+     '#default_value' => variable_get('fundraiser_standalone_cron_enabled', 0),
+   );
+
+  // Create series on cron option.
+  $form['fundraiser_sustainers']['fundraiser_cron']['fundraiser_sustainers_create_series_cron'] = array(
```
dbarbar: I think we should consider not making this an option. If creating the sustainers asynchronously gets us the user facing performance boost we're expecting, then I don't think we should let admins have this choice. That or we make it a hidden feature flag so that we can change it but admins can't. The long term goal should be to phase out the old way instead of supporting both ways. Having this option means we would have to support and maintain both.

epicflux: I think you're right that we should aim not to support both methods, but I think we'll have to initially support both methods to ease the transition. In the release of this function we'll let admin's know they should enable the cron creation option, and that the delayed creation process will become the de facto method in the next major release.

PR #759, opened 2015-06-16
Add Void Status for Commerce Orders

epicflux:
This adds a new status of Void available to commerce orders.


pcave:
I tried this out and it doesn't appear to impact anything. I also tried adding a corresponding Salesforce stage and mapped the new void status.

pcave:
void

PR #760, opened 2015-06-23
make email required for message action test mode

kljr:
Require an email address when activating test_mode. Update the description text on test_mode fields. Only send test_mode flag to API server when test mode is ture.


kljr:
Done.

PR #763, opened 2015-06-30
Sustainer Processing Improvements

epicflux:
@dbarbar I made some changes to the test to get the sustainers processing.

Initially the sustainers needed a sustainer key.

But, then I found it impossible to process both the success and failure cron runs in the same test, so I broke out the success and failure tests into separate classes and now they both complete.


dbarbar:
@epicflux The new tests look good to me and pass locally.

PR #767, opened 2015-07-16
Remove the comma from a number_format() for the donation amount.

dbarbar:


dbarbar:
@epicflux Ultimately, we should explicitly distinguish between the display value and the actual value. Tokens should use the formatted value. But then everything that uses that token needs to be revised to remove the hardcoded dollar sign.

It seems to be the case that the original amount value didn't have padded decimal places before I added the number_format(). I think eventually the number_format shouldn't be there at all.

PR #774, opened 2015-07-31
Donation Amount Not Changed in a Donation Update

epicflux:
Fix:

This sets the donation->amount to the new amount so fundraiser_donation table is updated.


epicflux:
The fix in commit 88ee661739e409b817e2a37969e09aa4f7b67c20 seems to have gotten into the main branch through this commit: 6c0ff97363d2660e7168beda6dc389a2bda1e181

PR #778, opened 2015-08-06
Fundraiser Offline Fixes to Handle Ajax and Multi-level Form Values

epicflux:
Fixes:
* Offline form customizations persist through ajax form submits.
* Mail field value creation avoids issues with multi-level form elements.


pcave:
Merged via pull request 899.

PR #779, opened 2015-08-06
Create parseable donation log entry for splunk reporting and aggregation 

pcave:
Adding an improved log message after each donation. The message is formatted in an structure ideal for Splunk.


fundraiser/fundraiser.module
```diff
@@ -3470,3 +3483,60 @@ function fundraiser_springboard_admin_admin_menu_items_alter(&$items) {
...
+
+  // Calculate the execution time.
+  $exec_time = round($log['end'] - $log['start'], 4);
+
+  // Setup the gateway details, the gateway name and config (rule name) are contained in the id.
+  list($gateway, $gateway_config) = explode('|', $donation->gateway['id']);
+
+  $pieces = array(
+    'datetime',
+    'category',
```
pcave: What's the purpose of the category piece?

fundraiser/fundraiser.module
```diff
@@ -3470,3 +3483,60 @@ function fundraiser_springboard_admin_admin_menu_items_alter(&$items) {
...
+ * Log that a donation has processed and some other details for Log Analysis.
+ *
+ * @param object $donation
+ *   The donation object.
+ * @param array $log
+ *   Additional log details, should include start/end times, and the type of transaction: charge, refund.
+ */
+function fundraiser_log_donation_processing($donation, $log = array()) {
+  // @todo Decide on an appropriate type for watchdog.
+  $type = 'fundraiser_donation';
```
pcave: With this fundraiser now has 5 types. I think grouping it with the normal 'fundraiser' type is probably ok.

fundraiser/fundraiser.module
```diff
@@ -3470,3 +3483,60 @@ function fundraiser_springboard_admin_admin_menu_items_alter(&$items) {
...
+    '%payment_gateway' => $gateway,
+    '%payment_gateway_config' => $gateway_config,
+    '%transaction_type' => $log['transaction_type'],
+    '%transaction_total_exec_time' => $exec_time,
+    '%donation_value' => $donation->donation['amount'],
+    '%donation_currency' => $donation->donation['currency']['code'],
+    '%country' => $donation->donation['country'],
+    '%origin_ip' => ip_address(),
+    '%success' => isset($donation->result['success']) ? $donation->result['success'] : 0,
+    '%reference_charge' => $donation->reference_charge,
```
pcave: In my initial donation $donation->reference_charge was empty thus resulting in a string of 13 pipe delimited values instead of 14. This probably needs an isset() like success has.

pcave:
Generally this looks pretty good. Does it need any sort of indicator for the site, or will that be available by other means?

epicflux:
The site identifier will be included in the full log message sent to splunk.

fundraiser/fundraiser.module
epicflux: This was requested by @danunezjr in the initial example of the message. While the value is a static 'INFO' now, the idea is that key could be used for other values in the future.

PR #787, opened 2015-08-24
Correct Usage of Thousand Separator and Multi-Currency in Refund Form

epicflux:
This code fixes the handling of the donation and refund amounts on the refund form to support multi-currency and + thousand amounts.

### DIFF
--- a/fundraiser/fundraiser.module
+++ b/fundraiser/fundraiser.module
@@ -2221,31 +2221,46 @@ function fundraiser_show_refund_form($did) {
  */
 function _fundraiser_refund_form($form, &$form_state, $did) {
   $form = array();
+
   // Given a donation ID, grab all relevant information to create the form.
   $donation = fundraiser_donation_get_donation($did);
-  $form['#donation'] = $form_state['#donation'] = $donation;
+  $form_state['donation_id'] = $donation->did;
+
   $form['payment_summary'] = array(
     '#type' => 'fieldset',
     '#title' => t('Payment Summary'),
     '#collapsible' => FALSE,
     '#collapsed' => FALSE,
   );
+
+  // Since we'll be doing math on these numbers we'll use the whole number values.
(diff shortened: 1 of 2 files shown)


fundraiser/fundraiser.module
```diff
@@ -2374,34 +2391,46 @@ function _fundraiser_refund_form($form, &$form_state, $did) {
...
   $refund_type = $form_state['values']['refund_type'];
+  $able_to_refund = $form_state['values']['able_to_refund'];
+
   if (empty($refund_type)) {
     form_set_error('refund_type', t('Select a type for the refund.'));
   }
   if ($refund_type != 'full') {
-    $amount = (float) $form_state['values']['amount'];
+    // Convert the entered amount to a whole number.
+    $amount = commerce_currency_decimal_to_amount(preg_replace('/[^\d\.]/i', '', $form_state['values']['amount']), $donation->currency);
```
pcave: We probably should have created a remove_commas() function or something similar instead of sprinkling this regex throughout the code.

fundraiser/fundraiser.module
```diff
@@ -2374,34 +2391,46 @@ function _fundraiser_refund_form($form, &$form_state, $did) {
...
+
     if ($amount <= 0) {
       form_set_error('amount', t('Enter a positive numerical refund amount.'));
     }
     else {
-      $refunded_total = (float) $form_state['values']['refunded_total'];
-      // Determine how much has been paid, less previous refunds
-      $able_to_refund = $donation->donation['amount'] - $refunded_total;
+      // Check if the amount set to refund is less than the amount left to refund.
       if ($amount > $able_to_refund) {
```
pcave: If refund amount is greater than the remaining balance the $refund_amount variable is never set and leads to an undefined variable on line 2432.

epicflux:
@pcave your suggested fixes have been applied.

PR #797, opened 2015-09-10
Adding admin menu alias for upsell preview to avoid use of admin UI

sheena-d:
Thanks to our admin menu path rewriting, the Upsell Preview tab on donation forms is currently redirected to /springboard/node/%/sustainer_upsell_preview, which causes the form's page wrapper to not be applied, thus making the Preview completely unhelpful for styling or preview of what the actually modal will look like. 

This patch creates a new entry in the admin menu aliases that makes the paths relevant to the upsell preview exempt from path rewrites.

Assembla ticket no. 596


sheena-d:
This is superseded by Pull Request #804 which is based on the correct branch.

PR #799, opened 2015-09-15
Adjusted expiry scheduling to support non-expiring, reusable tokens

jcestep:


pcave:
Instead of using jquery, you may want to look at using the Drupal #states system that was introduced in D7 with the FAPI. It can be used to hide/show etc, fields based on the state of another field. Seems like it might fit the bill here. Might be nice not having the jquery dependency here.

http://randyfay.com/states

jcestep:
I have replaced the form jQuery with Form API State code as requested.

PR #803, opened 2015-09-16
Fix improper assignent of form token variable

pcave:
Hotfix for users seeing "The form has become outdated" error message when creating target groups.


kljr:
drupal_valid_token() compares form_state['values']['form_token'] (the actual token) with a hash of $form['#token'] (which is usually the form_id, not an actual token).

$form_info['form_token'] = $form['#token'];

was causing form_state['values']['form_token']  to contain the value of $form['#token'] rather than the token itself, causing validation fail

PR #812, opened 2015-09-21
Set default sort to newest first for the sbv_donations view which is …

jcestep:
…defined in the springbaord_views module.


pcave:
I believe this was part of another branch. This code produces duplicates in the view when a date filter is applied.

pcave:
Confirmed fixed in pull request #822

PR #814, opened 2015-09-21
Update webform_confirmations and email_wrappers to load non-cached donation object

jesswalls:
Changes to:

* email_wrappers
* fundraiser_webform
* webform_confirmations to bypass cached donation object loading. This fixes an issue where some tokens are not replaced correctly in email wrappers.


pcave:
I'm reluctant to include this because doing a full donation reload is a database intensive process, and doing it in the context of the form submission may lead to performance issues. Another solution should be explored if possible.

PR #827, opened 2015-10-06
Updated webform settings form to remove the no redirect option when 

jcestep:
…webform_user is enabled.


webform_confirmations/webform_confirmations.module
```diff
@@ -263,6 +263,16 @@ function webform_confirmations_form_webform_configure_form_alter(&$form, &$form_
     '(See tokens below).');
   $form['submission']['redirection']['#description'] = t('Choose where to redirect the user upon successful submission. The Custom URL ' .
     'option supports Webform token replacements. (See tokens below).');
+
+  // Remove the "no redirect (reload current page)" option if webform_user is enabled, as this prevents user creation:
+  if (module_exists('webform_user')) {
```
pcave: This is going to remove the option from regular webforms as well. You also need to check if the current node is webform user enabled, like so:

```php
if (module_exists('webform_user') && _webform_user_is_webform_user_node($node)) {
  remove option to reload form...
}
```

But I'm also wondering if this should be in the webform_user module instead of webform_confirmations.

webform_confirmations/webform_confirmations.module
```diff
@@ -263,6 +263,16 @@ function webform_confirmations_form_webform_configure_form_alter(&$form, &$form_
...
+    unset($form['submission']['redirection']['redirect']['#options']['none']);
+    // If the current value of this field is set to "none" then reset it to "Confirmation page";
+    if ($form['submission']['redirection']['redirect']['#default_value'] == 'none') {
```
pcave: Let's not worry about existing forms. I think we've only seen this misconfigured once.

jcestep:
My latest commit, 6cb3ca3, moves the "remove no redirect option" code from webform_confirmation to webform_user and removes the code that resets the current value if it's already set to "no redirect" when webform_user is enabled.

Please let me know if further adjustments are needed.

PR #837, opened 2015-10-16
Prevent the sync'ing of donations with unmapped order statuses

pcave:
This is not complete. Still need to deal with probability and close date.


salesforce/salesforce_donation/salesforce_donation.module
```diff
@@ -634,6 +638,20 @@ function salesforce_donation_salesforce_genmap_map_fields_alter(&$fields, $conte
...
 
 /**
+ * Implements hook_salesforce_queue_batch_alter().
+ */
+function salesforce_donation_salesforce_queue_batch_alter(&$batch, SalesforceQueue $queue, SalesforceQueueProcessor $processor) {
+  if ($batch['object_type'] == 'Opportunity') {
+    // Nix any "StageName" tokens for unmapped payment stages.
+    foreach ($batch['items'] as $index => $item) {
+      if (!empty($item->sobject->fields['StageName']) && substr($item->sobject->fields['StageName'], 0, 7) == '[Stage:') {
+        unset($batch['items'][$index]->sobject->fields['StageName']);
```
pcave: So this just removes the stage field altogether? And this causes the opportunity to fail because stage is a required field. The problem with this solution is that with the stage removed it will be more difficult to figure out which order status has not been mapped correctly.

pcave:
Resolved in #1012

PR #848, opened 2015-10-26
Token filtering hotfix

kljr:
Removes extraneous token types and tokens from fundraiser-enabled nodes, and extraneous token types from message actions.


pcave:
The module will need to be enabled manually during upgrade.

PR #853, opened 2015-10-27
Added jQuery-based comma removal from the extra ticket amount field v…

jcestep:
…alue.


pcave:
Addressed server side in https://github.com/JacksonRiver/springboard_modules/pull/854.

PR #877, opened 2015-11-20
7.x 4.x 728 add submission delete hook to webform conf

jcestep:


kljr:
That sub-select could have an enormous result. Using hook_module_implements_alter to make webform_confirmations instance of hook_node_delete to fire before webforms instance, I think you could get the original SID before webform deletes it.

pcave:
Holding off on this for now. The submission delete will probably bomb on forms with hundreds of thousands of submissions.

pcave:
The data purge scripts will deal with these as well.

PR #879, opened 2015-11-23
Updated form code to add a validation function instead of replacing t…

jcestep:
…he array of validation functions.


pcave:
This approach did not pass QA when multiple payment method are enabled on a form. Work is ongoing.

pcave:
Invalidated by pull request 904.

PR #892, opened 2015-12-04
#490 Making help text changes and adding update hooks

juliakm:
Switching action taker/action takers/action-takers to
advocate/advocates. Changing elected official to legislator.
Adding update hooks. 

This pull request is against 7.x-4.x_social_advocacy and includes all of the documentation work. My understanding is that the initial docs work was merged into 7.x-4.x so please let me know if I need a separate pull request to complete adding the hooks to 7.x-4.x. 

The update hooks definitely need review.


kljr:
Hi Julia - 
There are conflicts in your patch, most likely the numbering of the message action update hooks.

juliakm:
I've cleaned up the errors. Please let me know if you still see issues. Thanks!

kljr:
Looks good, but I think update_hooks.patch file can be deleted.

juliakm:
That file is gone now.

kljr:
Ok. I see one more thing.

function sba_message_update_7002() {
  $field = field_info_instance('node', 'field_sba_subject_editable', 'sba_message');

should be:

  $field = field_info_instance('sba_message', 'field_sba_subject_editable', 'sba_message_action_message');

And that entire update hook should ideally live in sba_message_action.install, because the bundle creation logic now lives in that module.

juliakm:
Let me know if there's anything else to modify here.

kljr:
I've test this and it works correctly. However, now it needs to be requested against 7.x-4.x, as the social advocacy branch has already been merged into it.

PR #896, opened 2015-12-04
Check in submitted values (in addition to GET parameter) to see whether offline mode is set.

nadavoid:


nadavoid:
closing this PR in favor of the one that is for 7.x-4.x, #897.

PR #898, opened 2015-12-04
Cs 288 7.x 4.x sage offline email

nadavoid:
These are three cherry-picked commits that completed recent work regarding passing the email entered into the donation form, instead of the logged in user, to the relevant payment gateways.


kljr:
It looks like this change would cause the mechant id, the merchant key and application ID to be logged to watchdog along with the other request parameters, whereas previously it they were not. Not sure if that's something we want.

PR #899, opened 2015-12-04
Set form_state value with drupal_array_set_nested_value().

nadavoid:


nadavoid:
To test, submit a form in offline mode, with the email field configured two different ways. One, with the email field at the top level of the form. Second, with the email field inside a fieldset like Contact Information. When submitting the form, leave the email field empty. When viewing the webform submission in both cases, the email address should be the auto-generated one, such as 1452182288@sb-offline-donation.com.

PR #912, opened 2015-12-18
Performance improvements for page wrapper node operations

jesswalls:
On large sites with many contrib modules enabled and large amounts of cached data doing a full cache clear via drupal_flush_all_caches() after page wrapper node changes can cause performance issues. These changes allow bypassing drupal_flush_all_caches() after page wrapper node operations to limit the caches cleared and avoid rebuilding module data, etc.

* Adds configuration option to admin/config/content/page-wrappers to bypass Drupal's full cache clear during creation/cloning/updates to page wrapper nodes.
* Additional options for specifying contrib module cache tables to exclude from clearing.

### DIFF
--- a/page_wrappers/page_wrappers.admin.inc
+++ b/page_wrappers/page_wrappers.admin.inc
@@ -109,6 +109,35 @@ function page_wrappers_settings_form($form, $form_state) {
       '#default_value' => variable_get('page_wrappers_filter_by_og', FALSE),
     );
   }
+  
+  // Settings for caching on page wrapper node operations.
+  $form['page_wrappers_cache_settings'] = array(
+    '#type' => 'fieldset',
+    '#collapsible' => TRUE,
+    '#collapsed' => TRUE,
+    '#title' => t('Cache settings'),
+    '#tree' => FALSE,
+  );
+  $form['page_wrappers_cache_settings']['page_wrappers_skip_full_cache_clear'] = array(
+    '#type' => 'checkbox',
+    '#title' => t('Bypass drupal_flush_all_caches().'),
+    '#description' => t('Enabling this option will use a subset of drupal_flush_all_caches() routines during page wrapper node operations.'),
+    '#default_value' => variable_get('page_wrappers_skip_full_cache_clear', FALSE),
+  );
+  // Get a list of all potential contrib cache tables.
+  $hook_cache_tables = module_invoke_all('flush_caches');
+  $form['page_wrappers_cache_settings']['page_wrapper_cache_tables_exclude'] = array(
+    '#type' => 'checkboxes',
+    '#states' => array(
+      'visible' => array(
+        ':input[name="page_wrappers_skip_full_cache_clear"]' => array('checked' => TRUE),
+      ),
+    ),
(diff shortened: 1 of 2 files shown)


page_wrappers/page_wrappers.module
```diff
@@ -1310,3 +1310,49 @@ function page_wrappers_check() {
...
 }
+
+/**
+ * Flush caches on page wrapper node operations.
+ */
+function _page_wrappers_flush_caches() {
+  $skip_full_cache_clear = variable_get('page_wrappers_skip_full_cache_clear', FALSE);
+  // @see https://api.drupal.org/api/drupal/includes%21common.inc/function/drupal_flush_all_caches/7
+  // Below we skip rebuilding menu and module data, etc.
+  if ($skip_full_cache_clear) {
```
pcave: Is there any particular reason you made this configurable. Is there a reason why you want or need a full cache clear?

page_wrappers/page_wrappers.admin.inc
```diff
@@ -109,6 +109,35 @@ function page_wrappers_settings_form($form, $form_state) {
...
+  );
+  $form['page_wrappers_cache_settings']['page_wrappers_skip_full_cache_clear'] = array(
+    '#type' => 'checkbox',
+    '#title' => t('Bypass drupal_flush_all_caches().'),
+    '#description' => t('Enabling this option will use a subset of drupal_flush_all_caches() routines during page wrapper node operations.'),
+    '#default_value' => variable_get('page_wrappers_skip_full_cache_clear', FALSE),
+  );
+  // Get a list of all potential contrib cache tables.
+  $hook_cache_tables = module_invoke_all('flush_caches');
+  $form['page_wrappers_cache_settings']['page_wrapper_cache_tables_exclude'] = array(
```
pcave: Is there an ideal config for this setting?

page_wrappers/page_wrappers.module
jesswalls: Wanted an option to fallback to the original behavior in case of any issues presenting themselves since this was basically an urgent fix for a site already in production. No issues have been reported so far, so this could probably be changed to skip the full cache clear by default and remove that option.

page_wrappers/page_wrappers.admin.inc
jesswalls: It's hard to say since this depends on how the site is configured and what modules are enabled. I think cache_token should always be cleared, but wasn't sure about panels caching etc. Most of those tables should be safe to skip (e.g.  cache_salesforce_object, cache_views_data). In core every table implementing hook_flush_caches gets cleared, but I wanted to err on the side of being conservative.

The defaults used where this is in production are:

`array('cache_pulled_tweets', 'cache_rules', 'cache_views', 'cache_shorten', 'cache_salesforce_object', 'cache_clients', 'cache_libraries', 'cache_entity_og_membership', 'cache_entity_og_membership_type', 'cache_field', 'cache_image', 'cache_panels', 'cache_views_data');`

ruscoe:
Moved to https://github.com/JacksonRiver/springboard_modules/pull/913/ so the branch can exist in the JacksonRiver repo.

PR #920, opened 2015-12-30
7.x 4.x paypal payments

syntheticgoo:
PayPal checkout and refund functions for fundraiser donations.


syntheticgoo:
Added some fixes for when refund requests are initiated through PayPal dashboard, rather than SB.

PR #922, opened 2016-01-04
7.x 4.9 beta1 796 remove recent donors form layouts requirement

jcestep:


fundraiser/fundraiser.module
```diff
@@ -754,6 +754,18 @@ function _fundraiser_donation_settings_form(&$form, &$form_state) {
     '#type' => 'value',
     '#value' => isset($node->receipt_email_message) ? $node->receipt_email_message : '',
   );
+
+  // Share the "Display" fieldset with related modules (e.g., Form Layouts).
+  if (empty($form['springboard_display'])) {
```
pcave: I'm assuming the empty check is there because form layouts is also attempting to add it if empty?

pcave:
Manually merged.

PR #927, opened 2016-01-06
Prevents disabling of payment gateways 

scottatdrake:
....when they are in use on a donation form. Accomplished by disabling the submit button on the confirmation form and setting message(s) for the user to disable them on the donation form(s).

image

### DIFF
--- a/fundraiser/modules/fundraiser_commerce/fundraiser_commerce.module
+++ b/fundraiser/modules/fundraiser_commerce/fundraiser_commerce.module
@@ -2575,6 +2575,43 @@ function fundraiser_commerce_form_rules_ui_edit_element_alter(&$form, &$form_sta
 }
 
 /**
+ * Implements hook_form_FORM_ID_alter().
+ *
+ * Prevent disabling of a gateway if it is in use on a donation form by
+ * disabling the submit button on the confirmation form.
+ */
+function fundraiser_commerce_form_rules_ui_form_rules_config_confirm_op_alter(&$form, &$form_state) {
+  // Seems like there should be a better way to find the rule id than arg()
+  $rule = rules_config_load(arg(5));
+  // Take action only on rules provided by commerce_payment.
+  if ($rule->module == 'commerce_payment') {
+    // Select all records so we can search inside the serizalied gateways field.
+    $donation_forms = db_query("SELECT nid, gateways FROM {fundraiser}");
+    if ($donation_forms) {
+      while ($donation_form = $donation_forms->fetchAssoc()) {
+        $gateways = unserialize($donation_form['gateways']);
+        foreach ($gateways as $gateway) {
+          if (!empty($gateway['id'])) {
+            if (strpos($gateway['id'], $rule->name) && $gateway['status'] == 1) {
+              // Disable the submit button and create user message.
+              $form['actions']['submit']['#disabled'] = TRUE;
+              $node_title = db_query('SELECT title FROM {node} WHERE nid = :nid', array(':nid' => $donation_form['nid']))->fetchField();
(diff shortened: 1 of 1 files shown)


fundraiser/modules/fundraiser_commerce/fundraiser_commerce.module
```diff
@@ -2575,6 +2575,43 @@ function fundraiser_commerce_form_rules_ui_edit_element_alter(&$form, &$form_sta
...
 
 /**
+ * Implements hook_form_FORM_ID_alter().
+ *
+ * Prevent disabling of a gateway if it is in use on a donation form by
+ * disabling the submit button on the confirmation form.
+ */
+function fundraiser_commerce_form_rules_ui_form_rules_config_confirm_op_alter(&$form, &$form_state) {
+  // Seems like there should be a better way to find the rule id than arg()
+  $rule = rules_config_load(arg(5));
```
pcave: Not sure there is a more elegant solution for this. @kljr, are you aware of anything?

fundraiser/modules/fundraiser_commerce/fundraiser_commerce.module
```diff
@@ -2575,6 +2575,43 @@ function fundraiser_commerce_form_rules_ui_edit_element_alter(&$form, &$form_sta
...
+  // Take action only on rules provided by commerce_payment.
+  if ($rule->module == 'commerce_payment') {
+    // Select all records so we can search inside the serizalied gateways field.
+    $donation_forms = db_query("SELECT nid, gateways FROM {fundraiser}");
+    if ($donation_forms) {
+      while ($donation_form = $donation_forms->fetchAssoc()) {
+        $gateways = unserialize($donation_form['gateways']);
+        foreach ($gateways as $gateway) {
```
pcave: Low risk, but you might want to ensure $gateways is a valid array before iterating.

fundraiser/modules/fundraiser_commerce/fundraiser_commerce.module
```diff
@@ -2575,6 +2575,43 @@ function fundraiser_commerce_form_rules_ui_edit_element_alter(&$form, &$form_sta
...
+          if (!empty($gateway['id'])) {
+            if (strpos($gateway['id'], $rule->name) && $gateway['status'] == 1) {
+              // Disable the submit button and create user message.
+              $form['actions']['submit']['#disabled'] = TRUE;
+              $node_title = db_query('SELECT title FROM {node} WHERE nid = :nid', array(':nid' => $donation_form['nid']))->fetchField();
```
pcave: I'm generally not a huge fan of querying inside of a loop, but it's probably ok here.

pcave:
Great work. This looks really good! What do you think about changing the message up a bit to look like this:

The payment gateway _Example Payment_ cannot be disabled because it is currently in use on the following donation forms:

* Test Donation Form
* Scott's test donation form

fundraiser/modules/fundraiser_commerce/fundraiser_commerce.module
scottatdrake: Funny, I later found almost the exact same comment in commerce_authnet.inc: 
`$rule = rules_config_load(arg(5));  //@TODO: is there a better way to get the rule name?`

kljr: Yes, and that's Joel asking about my use of arg()...

fundraiser/modules/fundraiser_commerce/fundraiser_commerce.module
scottatdrake: Oh good call. Addressed in commit 594a3cb70bfc4c355569fb6a7f0e292e59e07815

scottatdrake:
@pcave, I've adjusted the display of the user message in commit 594a3cb70bfc4c355569fb6a7f0e292e59e07815

PR #930, opened 2016-01-08
re #794 - allow remote submissions to pass ip to webform

matt-thomas:
These changes allow the remote submission to pass "ip_address" as a parameter and have this address written to the webform submission.


matt-thomas:
I've updated this PR a bit. Commerce always uses ip_address() for hostname, so one needs to hack into the donation object and add it to the queue as an update.

springboard_api/resources/springboard_api.form_resources.inc
```diff
@@ -128,6 +136,25 @@ function springboard_api_form_action_submit($nid, $app_id = NULL, $submission =
...
+      if ($ip != ip_address()) {
+        // Update IP address on webform submission
+        $submission = webform_get_submission($nid, $sid);
+        $submission->remote_addr = $ip;
+        $sid = webform_submission_update($node, $submission);
+
+        // Update IP address on donation
+        if ($did = fundraiser_donation_get_did_by_sid($sid)) {
+          // Simulate the original donation event to rebuild and requeue.
+          if ($donation = fundraiser_donation_get_donation($did, TRUE)) {
```
pcave: I'm a little concerned about a full donation reload and update during the submission process. This is going to be a big bottleneck if the api is used heavily. 

Are you sure this works? For instance, the authorize.net module does it's own call to ip_address() in it's charge function.

PR #936, opened 2016-01-13
properly format signature before user_save

kljr:


pcave:
Minor change. Probably worthwhile to get into 4.9.

PR #937, opened 2016-01-13
move the personal campaign perms to pareant module so they get activated

kljr:
This will fix the issue of permissions not being created on feature activation due to install sequence of roles and permissions. A flush cache will restore any missing permissions on existing sites.


pcave:
Manually merged.

PR #943, opened 2016-01-21
Updated user/[uid]/recurring_overview to indicate if a recurring dona…

jcestep:
…tion series has been auto-cancelled.


pcave:
@jcestep, it might make more sense to do this in _fundraiser_sustainers_get_donation_sets_recurr_by_uid() since there is already a conditional that is checking for active and cancelled statuses.

PR #953, opened 2016-01-23
7.x 4.9 rc8 quicksign update hook

kljr:
I was under the impression that quicksign was not enabled on any sites, but yesterday was shown a quicksign form on a site. Here's a hook that will add the new webform component to any existing quicksign-enabled nodes. Of course, any custom layouts for existing quicksign nodes will need updating.


pcave:
There are no SB4 quick sign implementations out there so we don't need this.

PR #958, opened 2016-01-26
Adjusted opt-in checkbox/opt-in node data saving logic to work correc…

jcestep:
…tly when opt-in is enabled for new webform nodes.


kljr:
You can remove  $tmp_settings = array();

_fundraiser_recent_donors_add_opt_in_webform_component() is being called twice on every prexisting node re-save, once in the submit function and once in hook_node_update.

That function call and the call to _fundraiser_recent_donors_set_node_opt_in_data() could probably be confined to hook node update

jcestep:
Thanks for that feedback Kevin.  I have made those changes in my latest commit.

PR #964, opened 2016-02-18
7.x 4.x 877 webform user save

scottatdrake:
Prevents user_save from being called twice on webform submission.

This change takes advantage of the fact that user_save() can be used to both create and new user and update existing users. Rather than calling user_save() to create a user then immediately calling _webform_user_save_profile_map(), which calls user_save() again, we just pass _webform_user_save_profile_map() a bare $account object and let it create or update the user as needed.


scottatdrake:
Closing this in favor of a PR against 7.x-4.9.2-beta.

PR #983, opened 2016-02-26
7.x 4.x 889 advocacy success hook

kljr:
Create a success hook for actions.

This required changing the order in which submit hooks are executed for message actions, from the message action hook firing first, to firing after webform and webform user, so that the user account info is available when single-step message actions are submitted.

### DIFF
--- a/springboard_advocacy/springboard_advocacy.module
+++ b/springboard_advocacy/springboard_advocacy.module
@@ -1,4 +1,4 @@
- <?php
+<?php
 /**
  * @file
  * Base Springboard advocacy functions.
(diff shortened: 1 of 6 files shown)


springboard_advocacy/springboard_advocacy.module
```diff
@@ -901,3 +901,32 @@ function springboard_advocacy_update_multistep_components($sid, $node, $changed)
...
+ *
+ * With petitions, $data is the webform field values in a flat array.
+ *
+ * @param $node
+ * @param $contact
+ * @param $data
+ * @param $sid
+ */
+function springboard_advocacy_success($node, $contact, $data, $sid) {
+  module_invoke_all('action_success', $node, $contact, $data, $sid);
```
pcave: The hook should probably be namespaced. springboard_action_success or something like that.

PR #984, opened 2016-02-29
7.x 4.x.9.2 beta2 fix messages generated

kljr:
The reordering of the submit hooks on message actions for the success hook patch broke the "messages generated" count.

This restores the previous order, and adds a new, post-webform submit hook to call the success function from. In order to do that a new session variable had to be created to store the raw API response->messages object.

I wonder if for message actions and social actions, if the $data array in both cases should be the deliverable and undeliverable session variables, instead of the two different things they are now.

If that was the case, we wouldn't need the new session variable to store the raw response in this patch.


kljr:
I'm not sure why those extra commits are in the commit history for this pull request.

kljr:
Against wrong branch

PR #985, opened 2016-03-02
Added conditional that suppresses the modal if the donation amount mat…

jcestep:
…ches no bracket.


pcave:
This is merged manually into 7.x-4.10-beta1

PR #986, opened 2016-03-03
7.x 4.x p2p save donor name

jesswalls:
- Adds donor_name field to springboard_p2p_personal_campaign_action table.
- Allow other modules to implement hook_springboard_p2p_save_personal_campaign_action_alter().
- In springboard_p2p_fundraiser_donation_success() save p2p donor name (from profile fields) to  springboard_p2p_personal_campaign_action table.
- Adds a 'campaign-donors' menu tab to p2p personal campaigns, where a user with the permission 'Administer Springboard P2P' can edit the names shown on the donor roll.


springboard_p2p/springboard_p2p.module
```diff
@@ -2114,6 +2114,18 @@ function springboard_p2p_fundraiser_donation_success($donation) {
 
     // Now that the donation is successful, mark the action as visible.
     $action['status'] = 1;
+    
+    // If the donor has opted to show their name, save it with the action.
+    if ($action['show_name']) {
+      $account = user_load($action['uid']);
```
pcave: Do you know if this does a fresh user load from the db or does it get pulled out of cache?

pcave:
Can the update hook also go through each existing springboard_p2p_personal_campaign_action record, load the user, grab the name and store it?

springboard_p2p/springboard_p2p.module
jesswalls: Looks like this is not cached:
`entity_load('user', $uids, $conditions, $reset);`

so the only caching here would be the static cache in the Entity controller (default).

springboard_p2p/springboard_p2p.install
```diff
@@ -633,3 +640,52 @@ function springboard_p2p_update_7008() {
...
+    'not null' => FALSE,
+    'default' => NULL,   
+  );
+  
+  db_add_field('springboard_p2p_personal_campaign_action', 'donor_name', $donor_name);
+  
+  if (!isset($sandbox['progress'])) {
+    $sandbox['progress'] = 0;
+    $sandbox['current_sid'] = 0;
+    $query = "SELECT COUNT(sid) FROM {springboard_p2p_personal_campaign_action} WHERE show_name = 1 AND status = 1 AND action_type = 'fundraiser'";
```
jesswalls: @pcave Is this query to restrictive for updating the donor_name field?

springboard_p2p/springboard_p2p.module
pcave: Maybe we look at pulling the name out of the $donation object to avoid the user_load() call.

jesswalls: @pcave Do you have some more info on what you'd like to see here? I was assuming getting the $donation object via fundraiser_donation_get_donation(). From inspecting locally it looks like the loaded $donation object is larger than the loaded $user object.

pcave: I'm thinking about utilizing these 2 properties, which will always be the same.

`$action['donor_name'] = $donation->donation['first_name'] . ' ' . $donation_item['last_name'] = $donation->donation['last_name'];`

PR #996, opened 2016-03-07
re #258 - checking for existence of springboard_views 

sheena-d:
Added conditions to check that springboard_views module exists before invoking views defined therein.


sheena-d:
new pull request: https://github.com/JacksonRiver/springboard_modules/pull/1103

PR #1015, opened 2016-03-17
re #739 - set closedate if it's blank for pending transactions

matt-thomas:
When a pending transaction tries to sync with a blank closedate, a "required field missing" error is thrown. This change sets the blank closedate to the last changed date, day granularity only.

### DIFF
--- a/fundraiser/modules/fundraiser_commerce/fundraiser_commerce.module
+++ b/fundraiser/modules/fundraiser_commerce/fundraiser_commerce.module
@@ -950,6 +950,10 @@ function fundraiser_commerce_fundraiser_donation_get_donation($donation) {
       $donation->close_date = date('Y-m-d H:i:s', $donation->last_changed);
       $donation->transaction_date = date('Y-m-d H:i:s', $donation->last_changed);
     }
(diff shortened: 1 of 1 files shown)


fundraiser/modules/fundraiser_commerce/fundraiser_commerce.module
```diff
@@ -950,6 +950,13 @@ function fundraiser_commerce_fundraiser_donation_get_donation($donation) {
       $donation->close_date = date('Y-m-d H:i:s', $donation->last_changed);
       $donation->transaction_date = date('Y-m-d H:i:s', $donation->last_changed);
     }
+    elseif (!$donation->status_charged && !isset($donation->close_date) && !$recurring) {
+      $recurring = _fundraiser_sustainers_get_recurring_by_did($donation->did);
+      if (!$recurring) {
+        // Re #739 - set close date to last changed for pwmb
```
pcave: I'm really not thrilled baking payment method specific logic into this module. Is there a reasonable hook that can be used to do this in PWMB itself?

PR #1024, opened 2016-04-04
Add springboard_honeypot module to block IPs after N Honeypot rejections

benclark:
We don't want to use the built-in Honeypot trigger to immediately block all Honeypot rejections. However, we do want to block IPs that have multiple Honeypot rejections.

This module will block IPs after a configurable number of failures. The default is 50, which is high enough to target spammers using scripts/bots, but safe for other traffic.

Keeping the name of this module generic enough so we can expand on Honeypot integration down the road.

### DIFF
--- a/springboard_honeypot/springboard_honeypot.admin.inc
+++ b/springboard_honeypot/springboard_honeypot.admin.inc
@@ -1,0 +1,22 @@
+<?php
+
+/**
+ * @file
+ * Admin functions for springboard_honeypot module.
+ */
+
+/**
+ * Admin settings for springboard_honeypot.
+ */
+function springboard_honeypot_admin_form($form, &$form_state) {
+  $form['springboard_honeypot_count_threshold'] = array(
+    '#type' => 'textfield',
+    '#title' => t('Honeypot blacklist count threshold'),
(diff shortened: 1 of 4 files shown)


pcave:
This looks really good. How would we go about testing it?

benclark:
@pcave I have already deployed it to the QA site we set up for testing the SmartyStreets API hotfix. See my latest comment on ticket 444 in Hosted space.

As for testing, we would want to do the following:

1. Make sure springboard_honeypot is enabled.
2. Make sure honeypot is enabled on the form you wish to test.
3. Set the springboard_honeypot count threshold to something much lower than 50, for ease of testing. Something like 5 should work.
4. Set the honeypot timer threshold to something like 30 seconds (default is 5 seconds). That'll make it easier to trigger honeypot rejections.
5. Submit the form 5 times, each time within 30 seconds. You should see Honeypot reject the form submissions because you're submitting it "too quickly."
6. After the 5th submission, your IP will be blocked and you should not be able to load any pages.
7. To unblock yourself, you will need to go into the database directly and remove your IP from the `blocked_ips` table.

pcave:
This has also been manually merged into 7.x-4.10-beta1

PR #1025, opened 2016-04-05
Updates market_source_global_fields field description

scottatdrake:


scottatdrake:
This update to the field description was super trivial so I went ahead and merged it into 4.10-beta1 to get it on the 410 QA site.

PR #1033, opened 2016-04-07
fix php notice on donation add forms

kljr:
Need to add a check for the existence of an NID to prevent code execution on donation node/add form.


kljr:
Joel says this code is going away, so no need for a pull.

PR #1040, opened 2016-04-15
7.x 4.x data warehouse legacy content

scottatdrake:
@pcave and @CammoKing 

Could probably use a little more cleanup, but I think this is in a testable place. 

I merged in the beta1 branch a few days ago. Not sure if that is a problem or not. If needed, and can abandon this branch and cherry-pick my commits into a new one so the history and diff is clean.


scottatdrake:
Superseded by https://github.com/JacksonRiver/springboard_modules/pull/1121

PR #1041, opened 2016-04-18
Cleanup salesforce_log tables on cron run.

scottatdrake:


salesforce/salesforce_log/includes/salesforce_log.admin.inc
```diff
@@ -9,12 +9,12 @@
...
  */
 function salesforce_log_settings($form, $form_state) {
-
   $form['salesforce_log_batch_row_limit'] = array(
     '#type' => 'select',
     '#title' => t('Discard batch entries above the following row limit'),
-    '#default_value' => variable_get('salesforce_log_batch_row_limit', 10000),
+    '#default_value' => variable_get('salesforce_log_batch_row_limit', 0),
     '#options' => drupal_map_assoc(array(
+      '',
```
pcave: @scottatdrake, if an empty value gets saved into the salesforce_log_batch_row_limit variable, will it cause the query to error?

scottatdrake: @pcave, the query doesn't get run when salesforce_log_batch_row_limit has the value '' because we check the value first with `if ($row_limit > 0) {...run query...}`

PR #1042, opened 2016-04-19
7.x 4.10 beta1 976 button text

kljr:
Makes the Selected and Unselected image fields mutually required.

Fixes option text display issues with broken images

Moves image logic out of submit_text control structure.

Prevents form alter hook from firing on submission result edit page.


kljr:
Moving the logic out of the cntrol structure made the pull request hard tor ead. Looking at the individual commits is probably easier to understand.

kljr:
This not ready!

PR #1049, opened 2016-04-22
7.x 4.x data warehouse

CammoKing:


springboard_dw/springboard_dw.module
```diff
@@ -0,0 +1,1286 @@
...
+  $message_item['target_first_name'] = $message['targets']->first_name;
+  $message_item['target_last_name'] = $message['targets']->last_name;
+  $message_item['target_gender'] = $message['targets']->gender;
+  $message_item['target_state'] = $message['targets']->state;
+  $message_item['target_party'] = $message['targets']->party;
+  $message_item['target_district_code'] = $message['targets']->district_code;
+  $message_item['target_district_name'] = $message['targets']->district_name;
+  $message_item['target_role'] = $message['targets']->role;
+  $message_item['target_person_id'] = $message['targets']->person_id;
+  $message_item['created_at'] = date('U');
```
scottatdrake: These dates will be incorrect when we pass in old social actions to the warehouse. Perhaps we can fall back to the current date if the value isn't passed into this function.

PR #1056, opened 2016-04-27
Adding new hook for sustainer order insertion.

pcave:
In addition to the new hook, the queueing of the future donations has been moved into the salesforce_donation module via implementation of the new hook.

I tested both real-time and delayed sustainer creation and orders are being queued as expected.


pcave:
This change was required to make it possible for the orders to also be queued up for the data warehouse.

scottatdrake:
Looks good to me.

PR #1059, opened 2016-04-28
7.x 4.10 beta1 1017 paypal and authnet settings lost after save

kljr:
hook_form_rules_ui_edit_element_alter() will not work consistently when implemented in the fundraiser gateway include files. This patch renames the implementations in authnet.inc and paypal_wps.inc and calls them dynamically from fundraiser_commerce_form_rules_ui_edit_element_alter() instead, which is properly situation in a .module file.

It also fixes a fatal error that occurs when trying to access the payment method rule reaction settings from Rules UI menu paths instead of commerce payments UI menu paths.

The diff looks larger than it really is because of line indentation changes from the removal of two unecessary control structures wrapping the rest of the code in the hooks.

The following control structures were removed from the top of the hook implementations:

// This test is duplicate of code in Fundraiser_commerce_form_rules_ui_edit_element_alter
if (isset($form['parameter']['payment_method']['settings']['payment_method']['method_id']['#value'])) {
    $id = $form['parameter']['payment_method']['settings']['payment_method']['method_id']['#value'];

   // This test appears to be unecessary.
    $is_com = isset($form['parameter']['commerce_order']) ? 1 : 0;
    if ($id == 'authnet_aim' && $is_com)  {

### DIFF
--- a/fundraiser/modules/fundraiser_commerce/fundraiser_commerce.module
+++ b/fundraiser/modules/fundraiser_commerce/fundraiser_commerce.module
@@ -2697,8 +2697,15 @@ function fundraiser_commerce_form_rules_ui_edit_element_alter(&$form, &$form_sta
       ? $form['parameter']['payment_method']['settings']['payment_method']['method_id']['#value']
       : '';
     $method = commerce_payment_method_instance_load($method_id . '|commerce_payment_' . $method_id);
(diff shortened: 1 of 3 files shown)


kljr:
Perhaps this should go in a separate QA branch before being merged.

PR #1074, opened 2016-05-25
Fix "There are no Fundraiser settings to display" message

robertromore:
Don't display "There are no Fundraiser settings to display" when there actually are settings to display on the node creation form.


scottatdrake:
First commit! (balloon drop)

PR #1075, opened 2016-05-26
1034: Adding support for expiration behavior for fundraisers

robertromore:


fundraiser/fundraiser.info
```diff
@@ -9,6 +9,7 @@ dependencies[] = token
 dependencies[] = commerce
 dependencies[] = list
 dependencies[] = entity
+dependencies[] = date_popup
```
pcave: Is there any way around this new hard dependency? I know the date and related sub-modules are already included in our make file, but if there's a way to utilize the pop-up when it's available, and degrade to regular text field when it's not, that might be nice.

robertromore: Removed the hard dependency on the date_popup module. It now defaults to a text field and does date validation both client and server side.

robertromore:
Closing this in favor of https://github.com/JacksonRiver/springboard_modules/pull/1083

PR #1078, opened 2016-06-06
994: Implementing GA tracking for PayPal transaction events.

robertromore:
This PR adds a module, "springboard_measurement_protocol", which allows tracking PayPal events in GA by utilizing Google's Measurement Protocol service.

@pcave I had more work done, and I might commit it to another branch, but it involves abstracting some of the Measurement Protocol functionality into a different module and providing a class to construct and deliver a payload to the service. But I think this would be classified as "scope creep" so stopped and can work on it later if desired.

### DIFF
--- a/springboard_measurement_protocol/includes/springboard_measurement_protocol.admin.inc
+++ b/springboard_measurement_protocol/includes/springboard_measurement_protocol.admin.inc
@@ -1,0 +1,28 @@
+<?php
+
+/**
+ * @file
+ * Admin file for the Springboard Measurement Protocol module.
+ */
+
+/**
+ * Page callback for admin settings.
+ */
+function springboard_measurement_protocol_admin() {
+  if (module_exists('googleanalytics')) {
+    $form['paypal_events'] = array(
+      '#type' => 'fieldset',
+      '#collapsible' => TRUE,
+      '#collapsed' => FALSE,
+      '#title' => t('PayPal Events'),
+    );
+    $form['paypal_events']['springboard_measurement_protocol_paypal_enable'] = array(
+      '#type' => 'checkbox',
+      '#title' => t('Send events for PayPal transactions'),
+      '#description' => t("Send analytics to Google's Measurement Protocol service when a PayPal transaction occurs."),
+      '#default_value' => variable_get('springboard_measurement_protocol_paypal_enable', FALSE),
+    );
+  }
+
+  return system_settings_form($form);
+}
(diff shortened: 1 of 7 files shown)


springboard_measurement_protocol/springboard_measurement_protocol.module
```diff
@@ -0,0 +1,157 @@
...
+          // 48 bits for "node".
+          mt_rand(0, 0xffff), mt_rand(0, 0xffff), mt_rand(0, 0xffff)
+        );
+      }
+
+      $variables = array(
+        // The Measurement Protocol version.
+        'v' => 1,
+        // The Tracking ID.
+        'tid' => variable_get('googleanalytics_account', ''),
```
pcave: @robertromore, we need to remove the dependency on the contrib google analytics module. Can we add a custom setting to collect the ga account id on the measurement protocol's settings page?

springboard_measurement_protocol/springboard_measurement_protocol.module
```diff
@@ -0,0 +1,157 @@
...
+        $ch = curl_init();
+        curl_setopt($ch, CURLOPT_URL, $host);
+        curl_setopt($ch, CURLOPT_VERBOSE, 0);
+        curl_setopt($ch, CURLOPT_POST, 1);
+        curl_setopt($ch, CURLOPT_POSTFIELDS, $variables);
+        curl_setopt($ch, CURLOPT_RETURNTRANSFER, 1);
+        curl_setopt($ch, CURLOPT_NOPROGRESS, 1);
+        curl_setopt($ch, CURLOPT_FOLLOWLOCATION, 0);
+        curl_setopt($ch, CURLOPT_SSL_VERIFYPEER, 1);
+        curl_setopt($ch, CURLOPT_SSLVERSION, CURL_SSLVERSION_TLSv1);
```
pcave: @benclark, should we use CURL_SSLVERSION_TLSv1_2 here? Or perhaps CURL_SSLVERSION_DEFAULT and let the server figure it out?

springboard_measurement_protocol/includes/springboard_measurement_protocol.admin.inc
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * @file
+ * Admin file for the Springboard Measurement Protocol module.
+ */
+
+/**
+ * Page callback for admin settings.
+ */
+function springboard_measurement_protocol_admin() {
+  if (module_exists('googleanalytics')) {
```
pcave: We ned to remove the dependency on the google analytics module. It's going bye bye.

springboard_measurement_protocol/springboard_measurement_protocol.module
benclark: @pcave I would change it to CURL_SSLVERSION_TLSv1_2 and test it. If the gateway supports it, we should force the use of TLSv1.2. (Until something better comes along.)

PR #1085, opened 2016-06-10
#1033: Salesforce sync support for Card On File payment token GUIDs.

biwashingtonial:


fundraiser/modules/fundraiser_sustainers/fundraiser_sustainers.module
```diff
@@ -527,7 +535,7 @@ function fundraiser_sustainers_salesforce_genmap_map_fields_alter(&$fields, $con
         // @TODO: Add a configuration screen for specifying which fields get sent back
         // to Salesforce when a recurring donation charge is processed.
         $fields_to_update_on_charge = variable_get('fundraiser_sustainers_recurring_donation_charge_fields',
-          array('StageName', 'Amount', 'Gateway_Reference__c', 'Transaction_Date_Time__c', 'Probability'));
+          array('StageName', 'Amount', 'Gateway_Reference__c', 'Transaction_Date_Time__c', 'Probability', 'Card_On_File_Token__c'));
```
pcave: @biwashingtonial, I don't think we need to update the card on file field after a sustainer payment is made.

biwashingtonial: @pcave, this is per Jennie's request in this #1033 comment: https://app.assembla.com/spaces/springboard/tickets/1033/details?comment=977432113

"Note that I also think this means we should include this in the set of data that is sent down to Salesforce when a recurring donation is charged/failed"

pcave: This still isn't needed. When a user updates their billing information via the sustainer management page, all unprocessed donations are re-queued and updated. That is the point in which the new payment token would get exported to Salesforce. We can remove this.

PR #1087, opened 2016-06-14
Cs 288 7.x 4.x upsell auth only

nadavoid:
- Fixes sage cardonfile processing to respect the gateway's configured transaction type. e.g. auth only versus auth & capture.
- Allows modules to prevent upsell creating a sustainer series.


nadavoid:
There are 4 minor things reported by scrutinizer. I'll go ahead and fix those.

PR #1100, opened 2016-06-28
7.x 4.x 262 common fundraiser classes

sheena-d:
Adding field-specific classes to common fundraiser form elements

This is a replacement for the automatically closed pull request here: https://github.com/JacksonRiver/springboard_modules/pull/1063

### DIFF
--- a/fundraiser/modules/fundraiser_webform/fundraiser_webform.module
+++ b/fundraiser/modules/fundraiser_webform/fundraiser_webform.module
@@ -249,7 +249,7 @@ function fundraiser_webform_webform_component_render_alter(&$element, $component
     return;
   }
 
-  // Add the wrapper function to only certain component types.
+  // Do not add the wrapper to these component types.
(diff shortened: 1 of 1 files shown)


fundraiser/modules/fundraiser_webform/fundraiser_webform.module
```diff
@@ -235,6 +238,50 @@ function fundraiser_webform_webform_submission_presave($node, &$submission) {
 }
 
 /**
+ * Implements hook_webform_component_render_alter().
+ *
+ * Adds an additional wrapper function around webform elements.
+ */
+function fundraiser_webform_webform_component_render_alter(&$element, $component) {
+  // Only action on donation forms.
+  $node = node_load($component['nid']);
```
pcave: Are we guaranteed to always get a node here?

pcave:
This is going to be backed out because it is breaking existing sites.

PR #1101, opened 2016-06-28
7.x 4.x 981 remove fundraiser commerce gateway response table reroll

jcestep:
- Removes schema definition of {fundraiser_commerce_gateway_response}

- Adds install update to remove table {fundraiser_commerce_gateway_response} if it exists

- Removes reference to {fundraiser_commerce_gateway_response} in fundraiser_commerce.module

- Removes reference to {fundraiser_commerce_gateway_response} in commerce_authnet.inc


pcave:
I just realized we can't actually get rid of this table. Authorize.net allows additional response fields to be mapped to Salesforce, and that data is stored in this table.

PR #1112, opened 2016-07-05
Prevent multiple submissions when using enter key.

scottatdrake:


fundraiser/js/donation_validation.js
```diff
@@ -118,17 +118,24 @@
...
-              $(this).hide();
-              $('.fundraiser_submit_message').hide();
-              $(this).after('<div class="donation-processing-wrapper">' +
-                '<p class="donation-processing">Processing' +
-                '<span class="donation-processing-spinner"></span>' +
-                '</p>' +
-                '</div>');
+              if ($('.fundraiser-donation-form').data('submitted') === true) {
+                // Previously submitted, don't submit again.
+                event.preventDefault();
```
robertromore: You could also disable the button by adding `$(this).prop('disabled', true);`.

Also, this only works if the user has javascript enabled, which most users will have. Another line of defense would be using Drupal's locking mechanism, as explained [here](http://drupal.stackexchange.com/a/152181).

scottatdrake: Thanks for the review, @robertromore. I wasn't sure if disabling the submit button would also disable form submission when using the enter key. I also kind of wanted to stay away from altering the existing code that hides the button and replaces it with the "Processing...." button for fear that it would have downstream consequences on client sites that have customizations. 

I also saw that approach to server side locking, but lock_acquire() contains a db_insert() which seemed like a potential performance issue. I figure client-side prevention is probably good enough.

pcave:
Manually merged into the 7.x-4.11-beta branch.

PR #1125, opened 2016-07-08
7.x 4.11 beta 1061 message download

kljr:
extra white space and a bad regex mangled the download url.

presumably  preg_replace("@//@", "/", ...

was meant to remove double / at the end of the url, not in the "https://"


pcave:
Thanks @kljr!

PR #1129, opened 2016-07-15
Updated the donation report view to remove the date popup exposed fil…

jcestep:
…ter from when using the data export filter and, as a workaround, apply the data filtering by query alter vs exposed filter.


pcave:
@kljr, can you please review this?

kljr:
This works, although it uses mysql-specific functions to manipulate the dates. The database-agnostic equivalent of these three functions are available in date_api_sql.inc.

However, I'm not sure why the dates need to be formatted at all, and why you couldn't compare the timestamps instead.

kljr:
RIght, nevermind the "however"

PR #1137, opened 2016-07-21
7.x 4.11 rc7 yaml update

kljr:
adds test for environment var to limit which branches acceptance tests and simpletests are run on.

fixes a regex which was incorrectly finding legit functions with the string "debug"

updates php version


kljr:
having port issues on the acceptance tests still

PR #1147, opened 2016-08-09
1083: Prevent canceled donations from being charged in date mode

syntheticgoo:


fundraiser/modules/fundraiser_sustainers/modules/fundraiser_date_mode/fundraiser_date_mode.module
```diff
@@ -601,16 +601,22 @@ function fundraiser_date_mode_process_queue_item($item) {
...
+        fundraiser_sustainers_process_single_recurring_donation($log, $recurring, $sustainer_key);
+      }
+      catch (Exception $e) {
+        fundraiser_sustainers_processing_exception($e, $recurring);
+      }
     }
-    catch (Exception $e) {
-      fundraiser_sustainers_processing_exception($e, $recurring);
+    else {
+      watchdog('fundraiser_sustainers_date_mode', 'Sustainer order id !did was previously canceled, and has been skipped.', array('!did' => $recurring->did), WATCHDOG_NOTICE);
```
pcave: If we're only doing a log entry does that mean the record will stayed queued indefinitely?

PR #1149, opened 2016-08-10
Quick donate HMAC via logged in user

pcave:
This pull request makes the necessary changes to support QD HMAC actions via logged in user. This assumes the user account is still in the system.


fundraiser/modules/fundraiser_quick_donate/modules/fundraiser_quick_donate_springboard_hmac/fundraiser_quick_donate_springboard_hmac.module
```diff
@@ -48,7 +48,6 @@ function fundraiser_quick_donate_springboard_hmac_create_link($payload, $rollbac
     $time,
     'qd',
     $payload['card_id'],
-    $payload['gateway_id'],
```
pcave: We decided to simplify the token payload and only include card id.

fundraiser/modules/fundraiser_quick_donate/modules/fundraiser_quick_donate_springboard_hmac/fundraiser_quick_donate_springboard_hmac.module
```diff
@@ -79,75 +78,53 @@ function fundraiser_quick_donate_springboard_hmac_springboard_hmac_success($payl
     return;
   }
```
pcave: Lots of changes in fundraiser_quick_donate_springboard_hmac_springboard_hmac_success(). We really only need to see if the user account exists and that the payload user id matches the cookie uid. The last check is to ensure we have a valid form to redirect to. If all those things are good we log in the user and perform the redirect. The form will then just work as expected. No need to stash things in session or perform form population.

fundraiser/modules/fundraiser_quick_donate/modules/fundraiser_quick_donate_springboard_hmac/fundraiser_quick_donate_springboard_hmac.module
```diff
@@ -79,75 +78,53 @@ function fundraiser_quick_donate_springboard_hmac_springboard_hmac_success($payl
...
-  // Move quickdonate values into their own array for our own sanity.
   else {
-    foreach ($mapping as $key) {
-      $_SESSION['springboard_hmac']['quickdonate'][$key] = $payload[$key];
-      unset($_SESSION['springboard_hmac'][$key]);
-    }
+    // User no longer exists.
+    // TODO: This is were we would need the Springboard cookie to intervene
+    // and do the form population. But I'm not sold it's worth it as this point.
   }
```
pcave: Here is where we really need to decide what to do. If the user account no longer exists should we just redirect to the form or try some prepopulation magic value cookie values?

fundraiser/modules/fundraiser_quick_donate/modules/fundraiser_quick_donate_springboard_hmac/includes/fundraiser_quick_donate_springboard_hmac.admin.inc
```diff
@@ -51,29 +51,6 @@ function fundraiser_quick_donate_springboard_hmac_token_generator($form, $form_s
...
-  }
-
-  $form['fundraiser_quick_donate_springboard_hmac']['gateway_id'] = array(
-    '#type' => 'select',
-    '#title' => t('Gateway ID'),
-    '#description' => t('The gateway ID.'),
-    '#options' => $gateways,
-    '#default_value' => $defaults['gateway_id'],
-  );
-
```
pcave: Don't need the gateway_id any more and just relying on card_id instead.

fundraiser/modules/fundraiser_quick_donate/modules/fundraiser_quick_donate_springboard_hmac/fundraiser_quick_donate_springboard_hmac.module
```diff
@@ -162,33 +139,97 @@ function fundraiser_quick_donate_springboard_hmac_springboard_hmac_success($payl
...
+ *   The uid to verify.
+ * 
+ * @return boolean
+ *   True if the user is opted into quick donate. Otherwise false.
+ */
+function fundraiser_quick_donate_user_is_opted_in($uid) {
+  $cookie = springboard_cookie_get_cookie();
+
+  // Ensure uids match and optin date is set.
+  if ($cookie['uid'] == $uid) {
```
robertromore: This condition should check if the 'uid' value in $cookie is set first, otherwise PHP will throw a notice.

fundraiser/modules/fundraiser_quick_donate/modules/fundraiser_quick_donate_springboard_hmac/fundraiser_quick_donate_springboard_hmac.module
```diff
@@ -79,75 +78,53 @@ function fundraiser_quick_donate_springboard_hmac_springboard_hmac_success($payl
...
-      'status' => FALSE,
-      'reason' => FUNDRAISER_QUICK_DONATE_SPRINGBOARD_HMAC_INVALID_NODE_TYPE,
-    );
-    _fundraiser_quick_donate_springboard_hmac_redirect_to_default_node($nid);
+  // Make payload variable easier to work with.
+  $payload = _fundraiser_quick_donate_springboard_hmac_hmac_mapping($payload);
+  
+  // First, verify the uid in the payload matches what's stored in the cookie, if 
+  // it doesn't we can just stop right here.
+  if (!fundraiser_quick_donate_user_is_opted_in($payload['uid'])) {
```
robertromore: If 'uid' isn't a key in $payload, PHP will throw a notice.

fundraiser/modules/fundraiser_quick_donate/modules/fundraiser_quick_donate_springboard_hmac/fundraiser_quick_donate_springboard_hmac.module
```diff
@@ -162,33 +139,97 @@ function fundraiser_quick_donate_springboard_hmac_springboard_hmac_success($payl
  */
 function _fundraiser_quick_donate_springboard_hmac_get_nid($payload) {
   $parameters = drupal_get_query_parameters();
-  $mapping = _fundraiser_quick_donate_springboard_hmac_hmac_mapping();
-  $gateway_id = $payload[$mapping['gateway_id']];
-  $default_data = variable_get(FUNDRAISER_QUICK_DONATE_DEFAULT_FORM_PREFIX . $gateway_id, array());
+  
+  $card_on_file = commerce_cardonfile_load($payload['card_id']);
+  $gateway_id= $card_on_file->instance_id;
```
robertromore: Missing space between $gateway_id and =

fundraiser/modules/fundraiser_quick_donate/modules/fundraiser_quick_donate_springboard_hmac/fundraiser_quick_donate_springboard_hmac.module
```diff
@@ -162,33 +139,97 @@ function fundraiser_quick_donate_springboard_hmac_springboard_hmac_success($payl
...
+
```
robertromore: I think it'd be better to check for the form_id parameter first and return it, and then continue on with the rest of the conditions and returning as soon as possible, instead of saving the value to a variable to return at the end of the function. So it could be something like:

```
if (isset($parameters['form_id']) && is_numberic($parameters['form_id']) && fundraiser_quick_donate_form_has_gateway($paramters['form_id'], $gateway_id)) {
  return $parameters['form_id'];
}
if ($nid = fundraiser_quick_donate_get_default_form_for_gateway($gateway_id)) {
  return $nid;
}
return FALSE;
```

fundraiser/modules/fundraiser_quick_donate/modules/fundraiser_quick_donate_springboard_hmac/fundraiser_quick_donate_springboard_hmac.module
```diff
@@ -162,33 +139,97 @@ function fundraiser_quick_donate_springboard_hmac_springboard_hmac_success($payl
...
+ *
+ * @return int|FALSE
+ *   The nid of the default form for the gateway instance or FALSE.
+ */
+function fundraiser_quick_donate_get_default_form_for_gateway($gateway_id) {
+  $default_data = variable_get(FUNDRAISER_QUICK_DONATE_DEFAULT_FORM_PREFIX . $gateway_id, array());    
+  if (!empty($default_data)) {
     $default_data = unserialize($default_data);
-    return $default_data['nid'];
+    return $default_data['nid'];
```
robertromore: Extra whitespace at end of line should be removed.

fundraiser/modules/fundraiser_quick_donate/modules/fundraiser_quick_donate_springboard_hmac/fundraiser_quick_donate_springboard_hmac.module
```diff
@@ -162,33 +139,97 @@ function fundraiser_quick_donate_springboard_hmac_springboard_hmac_success($payl
...
+  }
+
+  return FALSE;
+}
+
+/**
+ * Helper function to redirect to a node if $nid > 0.
+ */
+function _fundraiser_quick_donate_springboard_hmac_redirect_to_form($nid = 0) {
+  drupal_goto('node/' . $nid);
```
robertromore: If this function is just calling another function, you can probably replace them. I had created this function to redirect to a node because I wanted to also do some validation. Not sure if that wasn't merged or if you removed it.

fundraiser/modules/fundraiser_quick_donate/modules/fundraiser_quick_donate_springboard_hmac/fundraiser_quick_donate_springboard_hmac.module
```diff
@@ -197,43 +238,34 @@ function _fundraiser_quick_donate_springboard_hmac_redirect_to_default_node($nid
...
+    ));
   }
-
-  // Otherwise this is a failure condition. Output the reason if configured to
-  // do so.
-  if (variable_get('fundraiser_quick_donate_debug', FALSE)) {
-    drupal_set_message($reason, 'warning', FALSE);
+  else {
+    // Otherwise, just take the user to the donation form.
+    drupal_goto(drupal_get_path_alias('node/' . $nid));
```
robertromore: Pretty sure this is my line of code, but we don't need the `drupal_get_path_alias` call.

fundraiser/modules/fundraiser_quick_donate/modules/fundraiser_quick_donate_springboard_hmac/fundraiser_quick_donate_springboard_hmac.module
```diff
@@ -197,43 +238,34 @@ function _fundraiser_quick_donate_springboard_hmac_redirect_to_default_node($nid
...
  * Implements hook_springboard_hmac_verify_payload().
  */
 function fundraiser_quick_donate_springboard_hmac_springboard_hmac_verify_payload($payload) {
-  $mapping = _fundraiser_quick_donate_springboard_hmac_hmac_mapping();
-  $card_id = $payload[$mapping['card_id']];
-  $gateway_id = $payload[$mapping['gateway_id']];
-  if (!isset($card_id) || !isset($gateway_id)) {
-    return t('No card or gateway information available.');
+  $payload = _fundraiser_quick_donate_springboard_hmac_hmac_mapping($payload);
+  $card_id = $payload['card_id'];
```
robertromore: Should we ensure `card_id` is set on $payload?

secure_prepopulate/springboard_cookie/springboard_cookie.module
```diff
@@ -109,6 +112,79 @@ function springboard_cookie_sb_social_share_event_save($share_data) {
...
+    }
+    if (array_key_exists('sbp_zip', $properties)) {
+      $profile_data['profile']['zip'] = $user_wrapper->sbp_zip->value();
+    }
+    if (array_key_exists('sbp_phone', $properties)) {
+      $profile_data['profile']['phone'] = $user_wrapper->sbp_phone->value();
+    }    
+  }
+  catch (EntityMetadataWrapperException $exc) {
+
```
robertromore: Should we log to watchdog if an exception is thrown?

secure_prepopulate/springboard_cookie/springboard_cookie.module
```diff
@@ -227,7 +303,7 @@ function springboard_cookie_tag_client($new_data = array()) {
         // If the existing and new values are arrays, simply combine them.
         $new_cookie[$name] = array_unique(array_merge($new_cookie[$name], $new_data[$name]));
       }
-      else if ($new_cookie[$name] != $new_data[$name]) {
+      else if (strtolower($new_cookie[$name]) != strtolower($new_data[$name])) {  // Non-case-sensitive comparison.
```
robertromore: `else if` should be `elseif`

fundraiser/modules/fundraiser_quick_donate/modules/fundraiser_quick_donate_springboard_hmac/fundraiser_quick_donate_springboard_hmac.module
robertromore: Also, this function could be written down to:
```
$cookie = springboard_cookie_get_cookie();
return isset($cookie['uid']) && $cookie['uid'] == $uid;
```

fundraiser/modules/fundraiser_quick_donate/modules/fundraiser_quick_donate_springboard_hmac/fundraiser_quick_donate_springboard_hmac.module
```diff
@@ -162,33 +139,93 @@ function fundraiser_quick_donate_springboard_hmac_springboard_hmac_success($payl
...
-  return drupal_deliver_page(MENU_NOT_FOUND);
+  return FALSE;
+}
+
+/**
+ * Determines if a given user is opted into quick donate.
+ *
+ * Uses the Springboard cookie to see if the given uid matches
+ * the value stored in the browser's cookie and that the quick
+ * doante opt-in data is set.
```
robertromore: "doante" should be "donate"

fundraiser/modules/fundraiser_quick_donate/modules/fundraiser_quick_donate_springboard_hmac/fundraiser_quick_donate_springboard_hmac.module
robertromore: I'm having trouble thinking of when the user account would no longer exist except in the case where an admin intentionally deleted their account. In that case, I think the user should be treated as a new user.

PR #1165, opened 2016-08-26
7.x 4.11.2 hotfix dw

CammoKing:
This update fixes PHP notices on the legacy dashboard, and maps donation charge advancements to update operations against an existing record in the data warehouse instead of creating a duplicate.

### DIFF
--- a/springboard_dw/springboard_dw.module
+++ b/springboard_dw/springboard_dw.module
@@ -281,19 +281,22 @@ function springboard_dw_fundraiser_donation_success($donation) {
   // Get the Drupal data warehouse queue.
   $queue = DrupalQueue::get("springboard_dw_export");
 
-  // Queue the donation object which should not be in the queue,
-  // so there's no need to check if this item is warehoused.
-  $donation_item = springboard_dw_map_donation_item($donation);
+  // Check if the item is warehoused already and run an update instead of
+  // creating a new record (mainly applies to sustainer charge advancements)
+  $is_donation_update = springboard_dw_item_is_warehoused('donation',
+    $donation->did) == TRUE ? TRUE:FALSE;
 
-  // Create the queue item.
+  // Map the donation object.
+  $donation_item = springboard_dw_map_donation_item($donation,
+    $is_donation_update);
+
+  // Queue the item or flag it for updating.
   $queue->createItem($donation_item);
 
   // Log the queue item.
   springboard_dw_log_queue_item($donation_item);
 
-  springboard_dw_track_item_warehouse_queued('donation', $donation->did);
-
-  // Check if the transaction id exists
+  // Check if the transaction id exists and track payment if so.
   if (isset($donation_item['transaction_id'])) {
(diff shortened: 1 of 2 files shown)


springboard_dw/springboard_dw.module
```diff
@@ -281,19 +281,29 @@ function springboard_dw_fundraiser_donation_success($donation) {
   // Get the Drupal data warehouse queue.
   $queue = DrupalQueue::get("springboard_dw_export");
 
-  // Queue the donation object which should not be in the queue,
-  // so there's no need to check if this item is warehoused.
-  $donation_item = springboard_dw_map_donation_item($donation);
+  // Check if the item is warehoused already and run an update instead of
+  // creating a new record (mainly applies to sustainer charge advancements)
+  $is_donation_update = springboard_dw_item_is_warehoused('donation',
```
pcave: This looks odd. Why not just

`$is_donation_update = springboard_dw_item_is_warehoused('donation', $donation->did);`

springboard_dw/springboard_dw.module
```diff
@@ -1244,6 +1264,29 @@ function springboard_dw_map_donation_item($donation, $update = FALSE) {
 }
 
 /**
+ * Map a charge advance item to a donation update.
+ * @param $donation
+ * @return array
+ */
+function springboard_dw_map_donation_charge_advance_item($donation) {
```
pcave: The name 'advance item' isn't entirely accurate in terms of what's actually happening. The advance charge feature Julia showed you is just a testing feature for us. If we can't advance the charge date of a sustainer we'd have to wait around a full month to test sustainer charges.

That said, this could probably use a better name.

springboard_dw/springboard_dw.module
```diff
@@ -281,19 +281,29 @@ function springboard_dw_fundraiser_donation_success($donation) {
...
+  $is_donation_update = springboard_dw_item_is_warehoused('donation',
+    $donation->did) == TRUE ? TRUE:FALSE;
+
+  // Map the donation object
+  if(!$is_donation_update) {
+    $donation_item = springboard_dw_map_donation_item($donation,
+      $is_donation_update);
+  }
+  else {
+    $donation_item = springboard_dw_map_donation_charge_advance_item
```
pcave: Currently springboard_dw_map_donation_charge_advance_item only takes one argument. This appears to be passing 2.

springboard_dw/springboard_dw.module
CammoKing: This is updated.

springboard_dw/springboard_dw.module
CammoKing: This method has been renamed to "springboard_dw_map_donation_sustainer_item".

springboard_dw/springboard_dw.module
CammoKing: This is updated also.

PR #1166, opened 2016-08-31
Store card_id in fundraiser donation

pcave:
This update stores the card_id that was stored after a successful donation in the fundraiser_donation table. This will make it possible to query which stored card is attached to a sustainer series.

TODO
* There currently isn't an upgrade path. Ideally the update hook would load all existing donations, extract the card_id from $donation->data and store it in the new location. This will require a batched operation because it will require loading every donation.

### DIFF
--- a/fundraiser/fundraiser.install
+++ b/fundraiser/fundraiser.install
@@ -184,13 +184,6 @@ function fundraiser_schema() {
       'nid_uid' => array('nid', 'uid'),
     ),
   );
-  if (module_exists('commerce_cardonfile')) {
-    $schema['fundraiser_donation']['fields']['card_id'] = array(
-      'description' => 'The cardonfile card ID used to process this donation',
-      'type' => 'int',
-    );
-    $schema['fundraiser_donation']['indexes']['card_id'] = array('card_id');
-  }
   $schema['fundraiser_refund'] = array(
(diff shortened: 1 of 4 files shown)


fundraiser/fundraiser.install
```diff
@@ -544,3 +544,15 @@ function fundraiser_update_7009() {
     }
   }
 }
+
+/**
+ * Add card_id field to fundraiser_donation.
+ */
+function fundraiser_update_7010() {
```
pcave: In addition to the update hook, the field needs to be added to the schema definition.

fundraiser/modules/fundraiser_commerce/fundraiser_commerce.module
```diff
@@ -2943,3 +2943,22 @@ function fundraiser_commerce_gateway_image_validate($element, &$form_state, $for
     form_error($element, t('Unselected image field is required'));
   }
 }
+
+/**
+ * Implements hook_fundraiser_donation_create_record_alter().
```
pcave: I don't think these alter hook implementations are needed anymore.

fundraiser/fundraiser.install
robertromore: Added the field definition back in the schema.

PR #1167, opened 2016-08-31
7.x 4.11.2 hotfix dw

CammoKing:
This hotfix creates a utility wrapper for logging that all data warehouse module logging gets funneled through so there is a centralized check for if watchdog logging is enabled.

### DIFF
--- a/springboard_dw/modules/springboard_dw_queue_legacy/springboard_dw_queue_legacy.drush.inc
+++ b/springboard_dw/modules/springboard_dw_queue_legacy/springboard_dw_queue_legacy.drush.inc
@@ -159,7 +159,7 @@ function springboard_dw_queue_legacy_setup_batch($type = '') {
     if (variable_get('springboard_dw_advocacy_authentication_token') == NULL
           || variable_get('springboard_dw_advocacy_api_endpoint') == NULL
       ) {
-      springboard_dw_watchdog_log('API endpoint and/or authentication token
+      watchdog('data_warehouse','API endpoint and/or authentication token
       have not been set.', NULL, WATCHDOG_ERROR, l(t('Fix this on the config page'),
         'admin/config/services/springboard-dw'));
       return drush_set_error(dt('API endpoint and/or authentication token have not been set.'));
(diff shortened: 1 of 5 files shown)


scottatdrake:
Looks good to me!

springboard_dw/modules/springboard_dw_queue_legacy/springboard_dw_queue_legacy.drush.inc
```diff
@@ -195,7 +195,7 @@ function springboard_dw_queue_legacy_setup_batch($type = '') {
   $record_count = count($records);
   drush_log(dt('@count records found to process.', array('@count' => $record_count)), 'ok');
   if ($record_count < 1 && !$idlist) {
-    watchdog("data_warehouse", 'Attempted to process %type records when there are :record_count remaining.', array('%type' => $type, ':record_count' => $record_count), WATCHDOG_WARNING);
+    springboard_dw_watchdog_log('Attempted to process %type records when there are :record_count remaining.', array('%type' => $type, ':record_count' => $record_count), WATCHDOG_WARNING);
```
pcave: I don't think this is going to work. springboard_dw_watchdog_log just expects a formatted string but it's being called here with parameters as if it were a direct call to watchdog()

springboard_dw/modules/springboard_dw_queue_legacy/springboard_dw_queue_legacy.drush.inc
```diff
@@ -159,7 +159,7 @@ function springboard_dw_queue_legacy_setup_batch($type = '') {
     if (variable_get('springboard_dw_advocacy_authentication_token') == NULL
           || variable_get('springboard_dw_advocacy_api_endpoint') == NULL
       ) {
-      watchdog('data_warehouse', 'API endpoint and/or authentication token
+      springboard_dw_watchdog_log('API endpoint and/or authentication token
```
pcave: Same problem as line 198

springboard_dw/modules/springboard_dw_queue_legacy/springboard_dw_queue_legacy.drush.inc
scottatdrake: Yeah, I was just saying to Cameron:

I think there is an argument to switching the two calls in the legacy module back to watchdog(). They are an error and a warning, not just informational. So, even if you UNcheck the “Enable watchdog logging of queue item processing” I think you still want them to log.

CammoKing: These are rolled back to watchdog() since we probably want these to log given their severity.

PR #1174, opened 2016-09-16
#204 Updating Youtube to YouTube

juliakm:
Updating text in P2P and Advocacy to have YouTube appear with a capitalized T instead of as Youtube.


juliakm:
Looks like some other stuff got in there from advocacy. Redoing

PR #1177, opened 2016-09-19
fix duplicate detection issue

kljr:
1. Create a target group with parameters "federal representatives | republican"
2. Search for targets with parameters Federal senators | Republican
3. Try to add an indivial senator to the recipients block.

Result: incorrect dupe target notice.

Cause: the 'Add' link is missing the 'role' paramter, so the duplication warning is triggered by the party parameter only.


kljr:
Closing temporarily because it causes a conflict with the groups branch

kljr:
Needs to go in 4.12, now that groups is delayed.

PR #1179, opened 2016-09-19
Hide upsell modal if the payment method is offsite recurring

nadavoid:


nadavoid:
@pcave Thanks for the suggestion. I like it. I've added that function and refactored the two conditions. How does it look to you?

nadavoid:
Closing in favor of #1184

PR #1181, opened 2016-09-20
Prevent duplicate upsell objects form exporting

pcave:
Adding a check to see if an upsell object has already been exported. This should prevent duplicates in Salesforce.


pcave:
Had another commit mixed in. Closing.

PR #1184, opened 2016-09-22
Hide upsell modal if the payment method is offsite recurring

nadavoid:
Originally #1179


nadavoid:
oh wait, I need to re-do my branch to not include those two unrelated commits.

PR #1202, opened 2016-09-28
fix default SOQL query and install form parameters

kljr:
Missing required param StageName in default SOQL query and install config function.


scottatdrake:
@pcave this all looks good to me. Thanks for the fix @kljr .

PR #1204, opened 2016-09-28
7.x 4.12 beta 1139 user role assign

kljr:
Original "assign user roles" patch was missing a form ID, and did not remove the administrator role from the possible role assignments.

Also adds a config form to Springboard Admin settings to select which user roles should be excluded from the assign user roles permission.

Uses the same approach user.module implements to disable "authenticated user" checkbox to disable disallowed role checkboxes.

related: https://github.com/JacksonRiver/springboard/pull/21


kljr:
Not ready

PR #1214, opened 2016-10-13
Updating ffg_image config to not restrict uploaded image size.

sheena-d:
Currently the configuration for the ffg_image field on fundraiser designation groups has a minimum and maximum image resolution setting. Since this setting is not configurable, it should be removed as to not restrict the possible front-end designs related to this feature.


sheena-d:
Added new line.

PR #1225, opened 2016-10-19
Updated the content type settings form to disable the fundraiser tick…

jcestep:
…ets enable/disable checkbox if fundraiser is currently enabled.


scottatdrake:
Merged into 7.x-4.14-beta

PR #1231, opened 2016-11-02
Data warehouse drush command should output a count of items processed

benclark:
Adds a counter of the raw number of items processed by the drush export queue command.

This will improve our Sumo Logic dashboard, as well as allow us to see how well the export queue API is performing so that we can dial-in the right `--limit` values for legacy data jobs.

I also threw in some tweaks to the `drush_log()` calls in the drush command callback.

### DIFF
--- a/springboard_dw/springboard_dw.drush.inc
+++ b/springboard_dw/springboard_dw.drush.inc
@@ -50,7 +50,10 @@ function drush_springboard_dw_export_queue_process() {
     drush_print('Processing the Springboard data warehouse export queue');
 
     // Process the items in the queue.
-    springboard_dw_process_queue();
+    $counter = springboard_dw_process_queue();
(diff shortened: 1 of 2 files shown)


springboard_dw/springboard_dw.drush.inc
```diff
@@ -50,7 +50,10 @@ function drush_springboard_dw_export_queue_process() {
     drush_print('Processing the Springboard data warehouse export queue');
 
     // Process the items in the queue.
-    springboard_dw_process_queue();
+    $counter = springboard_dw_process_queue();
+    if (!is_null($counter)) {
+      drush_log('Processed !counter items in Springboard data warehouse export queue.', 'ok');
```
benclark: Need to add `!counter` variable replacement here.

PR #1234, opened 2016-11-04
Drush command for sustainer processing

pcave:
Sustainers can now be processed via a drush command! The --uri passed in must match the configured sustainer key or processing will not happen. This pull request also adds the ability to pass a maximum number of sustainers to process.

`drush fsp --limit=50 --uri=https://7x.local `

### DIFF
--- a/fundraiser/modules/fundraiser_sustainers/fundraiser_sustainers.module
+++ b/fundraiser/modules/fundraiser_sustainers/fundraiser_sustainers.module
@@ -1191,20 +1191,19 @@ function fundraiser_sustainers_fundraiser_donation_delete($donation) {
 /**
  * Helper function, kick off recurring donations for payment. Called on cron usually.
  */
-function fundraiser_sustainers_process_recurring_donations() {
+function fundraiser_sustainers_process_recurring_donations($limit = 200) {
   // Provide a hook to allow for modules to respond.
   module_invoke_all('fundraiser_donation_recurring');
   $log = array(
     'successes' => 0,
     'fails' => 0,
   );
-  $donations = _fundraiser_sustainers_cron_get_recurring();
+  $donations = _fundraiser_sustainers_cron_get_recurring($limit);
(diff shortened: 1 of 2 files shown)


scottatdrake:
Great idea!

fundraiser/modules/fundraiser_sustainers/fundraiser_sustainers.module
```diff
@@ -4360,8 +4359,8 @@ function fundraiser_sustainers_preprocess_html(&$vars) {
  *   Part of the sustainer record for additional logging.
  */
 function fundraiser_sustainers_processing_exception(Exception $e, $sustainer) {
-  $message = 'Exception thrown while processing a sustainer. Sustainer ID: %did. %type: !message in %function (line %line of %file).';
-  $variables = array('%did' => $sustainer->did);
+  $message = 'Exception thrown while processing a sustainer - did=%did, gateway=%gateway, uid=%uid, %type=!message in %function (line %line of %file).';
```
benclark: Is `%type` meant to be replaced with a variable?

pcave: Yes. watchdog_exception() replaces it with details from the passed in exception.

benclark:
@pcave This looks good to me, although I haven't had an opportunity to run it locally. If you want to package it up and deploy it to the sustainergeddon site, I don't see any risks there. Once the code is up on web05, I can (or you can) manually add a Jenkins job that calls this drush command.

Another thought: since you're adding a way to set `$limit` on the parent function in `fundraiser_sustainers_process_recurring_donations()`, maybe you can have the default value be configurable with a Drupal variable? Something like:

```{php}
function fundraiser_sustainers_process_recurring_donations($limit = NULL) {
  if (!isset($limit)) {
    $limit = variable_get('fundraiser_sustainers_process_recurring_donations_limit', 200);
  }
  // ...
}
```

PR #1236, opened 2016-11-11
7.x 4.x designation update 50

sheena-d:
Updating Fundraiser Designation product type to include an "internal title" field. Also updating the management view to display this new field.


pcave:
Merged into 4.12.4

PR #1237, opened 2016-11-15
lock recurring doantion amounts on specific designation categories

kljr:
Make the donation amount field on line items uneditable if the donation product is in a specific category.


pcave:
Merged into 4.12.4.

PR #1240, opened 2016-11-29
t1495: Added Springboard cookie user ID check

biwashingtonial:
t1495: Added Springboard cookie user ID check function springboard_hmac_verify_user_cookie() and implemented it for login-and-redirect and quick-donate actions.

### DIFF
--- a/fundraiser/modules/fundraiser_quick_donate/includes/fundraiser_quick_donate.springboard_hmac.inc
+++ b/fundraiser/modules/fundraiser_quick_donate/includes/fundraiser_quick_donate.springboard_hmac.inc
@@ -277,6 +277,11 @@ function fundraiser_quick_donate_springboard_hmac_verify_payload($payload) {
     return;
   }
 
+  // Require a non-logged-in user to match the Springboard cookie.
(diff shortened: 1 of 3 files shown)


pcave:
These changes are slightly different than what we discussed. It looks like the individual hmac actions are performing the cookie check via hook_ springboard_hmac_verify_payload instead of a global check in springboard_hmac_verify_payload(). 

I thought we were going to do something like this at the bottom of springboard_hmac_verify_payload() in lieu of the commented out code.

`
// TODO: The check for sustainer upgrade is being done to prevent
// cookie checking for that action. In the future this may be
// removed so that the cookie check is performed for all implemented
// actions.
if ($parsed_payload['action'] != 'su') {
  return springboard_hmac_verify_user_cookie($payload);
}

`

biwashingtonial:
We certainly can do it that way. It feels cleaner to me for modules that add actions to opt in to the cookie check if they need it, than for springboard_hmac.module to have to keep a running list of actions that need to be opted out. Any idea whether future HMAC features are likely to have this cookie-based security requirement?

PR #1242, opened 2016-12-04
1457: Cancel a sustainer series when a stored card is deleted.

robertromore:


robertromore:
Thanks Matt, I've added some comments to my code. I've also given your comment to Kara for her thoughts and will update the PR if anything else needs to be changed.

robertromore:
From Kara:
"On #1, in the specific situation Matt notes - when the card for a recurring series is updated (vs. deleted) - we definitely would not want to cancel the rest of the series, we would want to change it over to the new card. I understand that technically in that scenario the original card is deleted, but it (and any others where the result is a valid, replacement card) is a different use case than the one this change is meant to address. The use case for this change is when the "delete" option is used from the stored cards page. I can't think of a scenario in which that shouldn't result in cancelation of a series associated with that card, since it leaves us with no way to process the future charges. If the change is being applied more broadly than that, we may indeed run into problems -- and thus I would prefer the change not be applied more broadly. Let me know if this is confusing and/or doesn't answer the question.

On #2, my strong preference is to not make this gateway-specific - if a card is deleted ( not updated ) , leaving us with no way to process future charges, the associated series should be canceled, there's no reason I can think of for that to vary by gateway. @phillip.cave do you have a different opinion? (On either of those.)"

So if the card is updated to a different card number, we do want to keep the existing series. But when it's explicitly deleted, the series should be removed.

PR #1246, opened 2016-12-12
Use email from user instead of order if offline

nadavoid:
Data is already saved as expected in Drupal. This change only affects the data sent to Authoriz.net when creating the CIM Profile.


nadavoid:
Don't merge this yet. I need to fix something first.

nadavoid:
Fixed. It's okay by me to merge this.

PR #1267, opened 2017-01-10
Fix payment nonce on cc and PayPal braintree

robertromore:


scottatdrake:
This looks good to me and fixes the issue locally.

PR #1270, opened 2017-01-17
Set TLS 1.2 explicitly for Braintree

scottatdrake:


braintree_integration/inc/braintree.inc
```diff
@@ -108,6 +108,7 @@ function braintree_initialize($settings) {
     Braintree_Configuration::publicKey($settings['public_key']);
     Braintree_Configuration::privateKey($settings['private_key']);
     Braintree_Configuration::environment($settings['environment']);
+    Braintree_Configuration::sslVersion(6);
```
pcave: Is there a constant that can be used instead of the number 6?

pcave: Nevermind. I don't think CURL_SSLVERSION_TLSv1_2 is available until PHP 7.

scottatdrake: Yep, I used that constant initially but then received a warning message about it when it got to the QA site.

pcave: Thanks. I love a good magic number!

PR #1274, opened 2017-01-23
filter by webform types on webform user and webform confirmation node…

kljr:
Don't run webform queries on node load if there are no webform types in the $types array.


scottatdrake:
I gave this a spin locally. It does what it says on the tin. Good catch, @kljr.

PR #1275, opened 2017-01-24
Implement submission delete hook to clean up leftover record

pcave:
The related record in webform_confirmations_submissions is not cleaned up when a submission is removed.


scottatdrake:
Tested locally. Works as advertised.

PR #1291, opened 2017-02-03
7.x 4.13 rc7 1213 awf p2p sync

kljr:
Fixes an issue with the webform countries field ajax request removing the form altered values set by springboard_p2p_form_alter() and adds a helper function to look up personal campaign ids. 

Also removes a non-functional code block that has been marked for removal for three years, but which was nonetheless calling node_load twice on personal campaign pages.


kljr:
The missing fields are of the type "value". If they had been regular hidden fields the ajax reload wouldn't have been a problem. I'm wondering if making them hidden might be the better solution.

PR #1324, opened 2017-03-02
Update which field is used to compute the “next charge” day for the u…

robertromore:
…sed for functionality on the stored cards page


fundraiser/modules/fundraiser_sustainers/fundraiser_sustainers.module
```diff
@@ -4634,10 +4634,19 @@ function fundraiser_sustainers_fundraiser_commerce_cardonfile_used_for($card_id)
...
+      // to the fundraiser_sustainers table which is set to "monthly" or
+      // "yearly"; or compute the difference between the last charge date and
+      // the next charge date and compare if it is less than a year or greater
+      // than a year. The former would be preferred since it's less
+      // computationally taxing.
+      '@type' => 'monthly',
       '@amount' => empty($result['amount']) ? '' : '$' . number_format($result['amount'], 2, '.', '') . ',',
       '@space' => empty($result['amount']) ? '' : ' ',
-      '@day' => date("jS", strtotime($result['npsp_next_payment_date'])),
+      '@day' => date("jS", strtotime($result['next_charge'])),
```
epicflux: @robertromore we don't need to run the next_charge value through strtotime since its a timestamp.

PR #1327, opened 2017-03-07
1682: Fixing issue with warning and wsod when editing a stored card a…

robertromore:
…nd viewing it again


epicflux:
This code for this PR is being handled in #1328 as well.

PR #1337, opened 2017-03-13
1577 enforce maximum donation amount

kljr:
Creates an upper limit of $999,000 on other amount field.


fundraiser/fundraiser.admin.inc
```diff
@@ -96,6 +105,15 @@ function fundraiser_admin_settings($form, &$form_state) {
   return system_settings_form($form);
 }
 
+function fundraiser_maximum_other_amount_validate($element, $form_state) {
+  if (!is_numeric($form_state['values']['fundraiser_maximum_other_amount'])) {
+    form_set_error('fundraiser_maximum_other_amount', t('Maximum "other amount" must be numeric, with no commas'));
+  }
+  if ($form_state['values']['fundraiser_maximum_other_amount'] > 999999.99) {
```
scottatdrake: @kljr, shouldn't this be `> variable_get('fundraiser_maximum_other_amount', 999999.99)` here? And in the error message just below.

fundraiser/fundraiser.admin.inc
```diff
@@ -109,8 +109,9 @@ function fundraiser_maximum_other_amount_validate($element, $form_state) {
   if (!is_numeric($form_state['values']['fundraiser_maximum_other_amount'])) {
     form_set_error('fundraiser_maximum_other_amount', t('Maximum "other amount" must be numeric, with no commas'));
   }
-  if ($form_state['values']['fundraiser_maximum_other_amount'] > 999999.99) {
-    form_set_error('fundraiser_maximum_other_amount', t('Maximum "other amount" must be less than 999999.99'));
+  $max_other_amount = variable_get('fundraiser_maximum_other_amount', 999999.99);
```
kljr: This was not an oversight. The maximum here needs to be hardcoded to 999999.99 to prevent an admin from entering a higher amount and thereby re-causing a pdo exception.

kljr:
note: need to rvert Scott's last commit.

scottatdrake:
@kljr I've reverted my previous commit. Thanks for the correction.

PR #1342, opened 2017-03-13
249 Market Source submodule for allowing overrides for offline Donations.

jesswalls:
This adds a submodule Offline Market Source that allows admins to be able to specify which market source fields are overrideable for offline donations

### DIFF
--- a/market_source/modules/offline_market_source/offline_market_source.module
+++ b/market_source/modules/offline_market_source/offline_market_source.module
@@ -1,0 +1,165 @@
+<?php
+
+/**
+ * @file
+ * Oflfline market source, provide a way for an administrative user to override market source fields
+ * on offline donation forms.
+ */
+
+/**
+ * Implements hook_module_implements_alter().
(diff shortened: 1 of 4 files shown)


market_source/modules/offline_market_source/offline_market_source.module
```diff
@@ -0,0 +1,165 @@
...
+            $form['submitted'][$key]['#type'] = 'textfield';
+            $form['submitted'][$key]['#weight'] = 0;
+          }
+        }
+      }
+    }
+  }
+}
+
+function offline_market_source_form_submit($form, &$form_state) {
```
epicflux: @jesswalls this submit function looks empty, maybe it's not needed and could be removed?

jesswalls: @epicflux It's been a while, so I don't recall the reason for that function. It can probably be removed. I can test locally and make that modification if you'd like.

PR #1343, opened 2017-03-13
7.x 4.x 1685 country state field consistency

jcestep:


fundraiser/modules/fundraiser_commerce/fundraiser_commerce.module
```diff
@@ -809,6 +809,10 @@ function _fundraiser_commerce_zone_field_display($form, $form_state, $field) {
   // Get countries.
   $countries = fundraiser_commerce_get_countries();
 
+  if (empty($form_state['values']['submitted']) && !empty($form_state['values'])) {
```
epicflux: @jcestep Please add a comment above this line, something like: Copying the form values into a 'submitted' array to work with the zone field display.

jcestep: Done.

PR #1352, opened 2017-03-16
1626: Set correct formId for generic braintree forms; refactor code t…

robertromore:
…hat updates braintree payment information when existing user uses existing payment method; fix token variable, use card’s unique ID to determine if/when to create/update stored cards.


epicflux:
@scottatdrake when you have the chance please review these changes.

PR #1357, opened 2017-03-21
Store gateway info using cache_set().

scottatdrake:
If I'm looking at new relic correctly, building all the gateway info can take 140-200ms. This caches it using Drupal's cache_set() and flushes the cache when any rule is inserted, deleted, edited.


fundraiser/fundraiser.module
```diff
@@ -2736,15 +2736,21 @@ function _fundraiser_get_field_fields($field_info = NULL, $fields = array()) {
...
+  $gateways = &drupal_static(__FUNCTION__);
+  if (!isset($gateways)) {
+    // Try to get gateway info from the cache.
+    if ($cache = cache_get('fundraiser_gateway_info')) {
+      $gateways = $cache->data;
+    }
+    else {
+      // If not, we go asking anyone else - usually the payment glue module.
+      $gateways = module_invoke_all('fundraiser_gateway_info');
+      drupal_alter('fundraiser_gateway_info', $gateways);
```
scottatdrake: Actually, this drupal_alter() poses a problem for caching gateway info because it can alter them based on any random thing. Eg fundraiser_offline_fundraiser_gateway_info_alter() alters gateway info based on the presence of a $_GET param. This might prevent caching gateway info for longer than just the current request.

PR #1364, opened 2017-03-23
Load commerce payment include to avoid missing function error.

epicflux:


epicflux:
This was a duplicate of #1336

PR #1373, opened 2017-03-28
7.x 4.15 beta performance tweaks

epicflux:


fundraiser/modules/fundraiser_commerce/fundraiser_commerce.module
```diff
@@ -1054,6 +1067,7 @@ function fundraiser_commerce_fundraiser_donation_create($donation) {
   $quantity = isset($donation->donation['quantity']) && !empty($donation->donation['quantity']) ? $donation->donation['quantity'] : 1;
 
   // Give other modules an opportunity to generate the order entity.
+  // @TODO nobody implements this. Can we get rid of it?
```
epicflux: @pcave  Answering this TODO: this function is used in custom code on client sites. I think its best to remove or rewrite the TODO.

pcave: Ok. Do we have to re-add the commerce_order_save() call as well? I think Scott figured out we save the order 8 times during a single request and was trying to eliminate any unnecessary calls.

scottatdrake: There is a commerce_order_save() further down in this same function. Taking this order_save out didn't appear to have any impact but perhaps clients do doing custom things here that I'm not aware of, perhaps in the hook_ fundraiser_commerce_generate_line_items() that follows.

epicflux: I haven't noticed any issues with removing the commerce_order_save().

epicflux:
Closing this in favor of #1413

PR #1376, opened 2017-03-28
prevent session storage after quicksign

kljr:
Creates a hook to allow other modules to disable webform_user session storage during a submission.

### DIFF
--- a/webform_user/webform_user.module
+++ b/webform_user/webform_user.module
@@ -571,7 +571,7 @@ function webform_user_webform_submit($form, &$form_state) {
       }
       webform_submission_update($node, $submission);
       // Save webform user fields in user session for the existing user.
-      if (_webform_user_is_webform_user_session_store()) {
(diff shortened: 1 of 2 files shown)


webform_user/webform_user.module
```diff
@@ -1297,8 +1297,22 @@ function _webform_user_is_webform_user_node_type($type) {
...
+ *
+ * @return bool
+ *   Whether to allow session storage.
  */
-function _webform_user_is_webform_user_session_store() {
+function _webform_user_is_webform_user_session_store($form_state) {
+  // Allow other modules to prevent webform user session storage.
+  $session_cancel = module_invoke_all('webform_user_session_storage_cancel', $form_state);
+  foreach ($session_cancel as $cancel) {
+    if (!$cancel) {
```
epicflux: @kljr using the identical operator here ($cancel === FALSE) would force functions to expressly deny the storage. If a function returned NULL it would not prevent the storage.

PR #1377, opened 2017-03-28
Reset card number and cvv values in form when navigating away from page

robertromore:


fundraiser/js/donation_validation.js
```diff
@@ -16,6 +16,17 @@
         ;
       }
 
+      var fields_to_reset = [
```
epicflux: @robertromore we won't be able to target specific ids as they can change. And we probably want to target fields beyond just credit.

There is a fieldset wrapper around payment the fields with a consistent class 'fundraiser-payment-fields'.

We should be safe to reset the value of any input within that fieldset.

fundraiser/js/donation_validation.js
```diff
@@ -16,6 +16,17 @@
...
+        '#edit-submitted-payment-information-payment-fields-credit-card-number',
+        '#edit-submitted-payment-information-payment-fields-credit-card-cvv'
+      ];
+      $(window).unload(function() {
+        $(fields_to_reset.join(',')).each(function(index, el) {
+          $(el).val('');
+        });
+        return "unloaded";
```
epicflux: @robertromore What is the purpose for this line: return "unloaded";?

pcave:
Confirmed the card number and cvv fields are being cleared when using the back button in Firefox.

PR #1378, opened 2017-03-28
7.x 4.x 1705 clear salesforce object cache

kljr:
Adds a menu item to the genmap and slaesforce entity forms to clear the cache.


pcave:
I like that you added a new module to handle this. Springboard administrators will need to access this feature so I'm wondering if they currently have the right permissions. I'll need to check on that.

PR #1379, opened 2017-03-29
7.x 4.x designations minimum amount

kljr:
Fundraiser designations allows fund id and donation amount to be passed via query parameters. If an invalid value is passed for amount (e.g., "$40" with a currency symbol), the amount in the form becomes "0". The zero amount then evades minimum amount detection during form submission. This patch applies minimum amount validation.


scottatdrake:
Closing this PR has it is superseded by the combined branch in PR #1439.

PR #1382, opened 2017-03-29
Move extra_props to top level.

scottatdrake:


scottatdrake:
@CammoKing has pointed out that this operation (making extra_props top level fields) is actually supposed to be happening on the receiving end of the API. He as a fix coming. Closing this PR.

PR #1397, opened 2017-04-04
1604 1611 1626 1630 - 4.15 braintree fixes

epicflux:


epicflux:
This branch contains a round-up of fixes for 1604, 1611, 1626, 1630.

PR #1398, opened 2017-04-04
7.x 4.x 1710 designation clone overrides

kljr:
Fixes:

1. Double serialization of fund group ask amounts was causing ask amounts to be lost when cloning.
2. Missing issets resulted in php notices when creating ask amount overrides on fund groups.
3. Incorrect view name was causing issues with draggable weights being ignored on the front end form.


scottatdrake:
Closing this PR has it is superseded by the combined branch in PR #1439.

PR #1403, opened 2017-04-06
7.x 4.x 1763 applepay compact flow

robertromore:


braintree_integration/modules/braintree_applepay/braintree_applepay.module
```diff
@@ -1,21 +1,24 @@
 <?php
+
 /**
  * @file
  * Module file for braintree_applepay.
  */
 
+define('BRAINTREE_APPLEPAY_PAYMENT_METHOD_ID', 'braintree_applepay|commerce_payment_braintree_applepay');
```
epicflux: @robertromore we can't expect the enitre payment method id to always be the same. However the first portion of the that string 'braintree_applepay' will always be the same. 

In the code that checks the payment method you can do this:

list($method_id, $rule_name) = explode('|', $method_instance['instance_id']);

to breakout the values.

braintree_integration/modules/braintree_payment/js/braintree_payment.paypal.admin.js
```diff
@@ -0,0 +1,47 @@
+(function ($) {
+  Drupal.behaviors.brainTreePayPalCompact = {
+    attach: function (context, settings) {
+      var errors = Drupal.settings.paypalCompactErrors || [];
+      //if (typeof(Drupal.settings.paypalCompactErrors) != 'undefined') {
```
scottatdrake: Should this be commented out? If so, it might be better to just delete the conditional.

PR #1404, opened 2017-04-07
re #1712 - replace all email validation with a single custom function

matt-thomas:
See ticket 1712 for history. Inconsistent email validation functions are used throughout Springboard. Some use filter_var, some use preg_match, and some modules are using core's valid_email_address(). For instance, email wrappers uses both its internal email validation function and core's valid_email_address(). We want to avoid the core function because it will validate email addresses without an extension in the domain. 

This PR replaces all the validation functions with a single helper function in springboard.module and shores up .info files so that springboard is always required for modules that need this function.


email_wrappers/email_wrappers.module
```diff
@@ -199,7 +211,11 @@ function _email_wrappers_node_edit_email_fields_validate($form, $form_state) {
 
     // Validate the field's value or values:
     foreach ($form_state['values'][$email_field][$lang] as $key => $email_value) {
+<<<<<<< HEAD
```
epicflux: @matt-thomas looks like a minor code conflict to resolve.

PR #1411, opened 2017-04-11
make springboard dashboard permissions finer-grained

kljr:
Springboard assumes if you have the “access dashboard” permission, you should be able to view all form types. That sounds reasonable, but it also does not check if you have the ability to create form types, and displays the create buttons for each type to everyone. This patch wraps the create buttons in an access check.

The marketing and analytics tab is also controlled by the “access dashboard” permission. The links on the marketing and analytics page are also included without any subsequent permissions check

So, in this case, the menu item for the marketing page needs to use a permission callback instead of checking dashboard perms. The permission callback needs to parse all of the user’s current permissions to see if he can access any one of the multiple marketing and analytics sub-pages.

And then, if the user has access to some or all marketing and analytics functions, the iterator that produces the links on the page must check each link for access.

The "reports" section uses the same page callback as the marketing section, and is also patched by this.


kljr:
merged into another branch

PR #1416, opened 2017-04-14
1069 Adds salesforce_merge_handler

scottatdrake:
When Contacts are merged in SF, one of them gets deleted, and we may end up with a Drupal user associated to a Contact that no longer exists. The next time the Drupal user is synced, SF returns a ENTITY_IS_DELETED error. 

This module looks for those errors and, when found, deletes the salesforce_sync_map record associated with it, and refreshes the sync item (as an UPSERT instead of UPDATE). The next time the salesforce sync process runs, the email should match the "other" Contact and we end up with a Drupal User associated with the right Contact.

### DIFF
--- a/salesforce/salesforce_merge_handler/salesforce_merge_handler.module
+++ b/salesforce/salesforce_merge_handler/salesforce_merge_handler.module
@@ -4,8 +4,8 @@
  * Implements hook_salesforce_sync_fail_item().
  *
  * Examines statusCode returned from failed sync items for ENTITY_IS_DELETED.
- * When found, the corresponding record from salesforce_sync_map is deleted and
- * the queue item is refreshed.
+ * When found, the corresponding Contact and Account records from
+ * salesforce_sync_map are deleted and the queue item is refreshed.
  *
  * When the queue item is synced, it should match on the email field and a new
  * salesforce_sync_map record created that referenced the new SF record.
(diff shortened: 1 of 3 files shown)


salesforce/salesforce_merge_handler/salesforce_merge_handler.module
```diff
@@ -0,0 +1,31 @@
...
+ * When found, the corresponding record from salesforce_sync_map is deleted and
+ * the queue item is refreshed.
+ *
+ * When the queue item is synced, it should match on the email field and a new
+ * salesforce_sync_map record created that referenced the new SF record.
+ */
+function salesforce_merge_handler_salesforce_sync_fail_item($item, $message, $result) {
+  if (isset($result->errors)) {
+    foreach ($result->errors as $error) {
+      if ($error->statusCode == 'ENTITY_IS_DELETED') {
```
pcave: Should this conditional also check the item type to see if it's a Contact? Otherwise this will run for donations, actions, etc.

pcave: $item->object_type

scottatdrake:
@pcave, do you think this should this delete the Account salesforce_sync_map record in addition to the Contact record? I don't know what a Salesforce merge doses with Accounts, do you?

salesforce/salesforce_merge_handler/salesforce_merge_handler.module
scottatdrake: Good call. I added a conditional for this at the top.

pcave:
@scottatdrake, good point. Please delete the account salesforce_sync_map record as well because the merged contact may or may not be associated with the same account. The system will grab the account id of the contact after it is synced up again.

scottatdrake:
@pcave I've adjusted this so that both the Account and Contact records are deleted.

PR #1418, opened 2017-04-14
7.x 4.x springboard perms combined update

kljr:
## Springboard Dashboard and Advocacy Dashboard permission and behavior updates.

These changes are independent of the new Springboard Groups implementation, but are required to maintain consistency both with and without Groups.

**Updates to “Springboard Admin Allowed Forms” settings and supporting code.**

This is the very well-hidden configuration which lives on the "springboard/settings/springboard" menu path, inside the collapsed "Springboard administration settings" Fieldset in the "Springboard Admin Allowed Forms" checkbox configuration subset.

1. Adds an alter hook that lets you remove content types from the list in “Springboard Admin Allowed Forms” settings so that a particular content type can be always excluded from the dashboard settings if desired. This new hook has been implemented in advocacy so that petitions, social actions and message actions do not appear as options in the setting.

2. Makes the dashboard panes respect the “Springboard Admin Allowed Forms” settings. Formerly the "Recent Donation Forms" pane and the "Recent Forms" pane disregarded this setting, while the "All Forms" pages the panes linked to obeyed it. This achieved by implementing the function springboard_admin_springboard_views_springboard_dashboard_pane_alter().

3. Adds "fundraiser types" to hook_springboard_views_springboard_dashboard_panes_alter(), so that they are alterable in that function just like the webform user types. This allowed the implementation in springboard_admin_springboard_views_springboard_dashboard_pane_alter() to remove fundraiser types from the recent donation forms pane in accordance with the setting, if such setting is configured.

**Updates to the Springboard dashboard page and supporting code:**

1. The Springboard dashboard panes will not display if a user does not have "edit any",  "create" or administrative access to any of *node types* that the pane displays. Previously the panes displayed even if the user could manage none of the content in them.

2. If a user does have edit or create access to *some* of the *node types* in the dashboard panes, the panes and the links to the full view will display, but the panes and pages will only show nodes of the *node types* that the user has edit or create access to.

In contrast, on a Springboard Group-enabled site, the dashboard will only list content that is in the selected group and to which the user has edit or create access.

3. Advocacy form types have been removed from the "Recent Forms" pane on the Springboard dashboard. This is because they have always been hard-coded to *not appear* on the corresponding "all forms" page that the "recent forms" page links to.

4. "Create" buttons on dashboard pages check whether a user can create that content type prior to displaying.

5. "Options" selectors on dashboard pages now check whether the user has create access for pagwrappers or email wrappers prior to displaying. @todo - perhaps this should be expanded to "edit any" as well

**Updates to the Advocacy dashboard and supporting code:**

1. Advocacy related hard-coded interventions have been moved out of the springboard views module and the springboard admin module into alter hooks in the advocacy modules themselves. The behavior isn't changing, just the location of the code that is producing the following behaviors: 

-  removing advocacy links from the "all forms" menu item, 
- removing advocacy content types from the "all forms" page (this is not strictly necessary anymore after #2 above, but is included for backward compatibilty in existing sites).

2. Formerly access to the Advocacy Dashboard and Advocacy top-level menu item was controlled by a check for whether the user had "create" access for any advocacy type (message action, social action, petition). This has been expanded to "create" or "edit any". @TODO Create a named permission for access to this menu item instead? "access advocacy dashboard" -  but its use would be limited to this one menu item.

3. The individual dashboard panes on "springboard/advocacy/actions" view all actions page have been wrapped in a new permissions check. If the user is not an admin or advocacy admin, or does not have "create" or "edit any" access to the content type displayed in the pane (petition, message action, social action), that dashboard pane will not appear.

4. Because it is conceivable (if unlikely) that a user might have "edit any" but not "create" permissions, the "create a new" buttons on the dashboard have been wrapped in a permissions check."

5. The original task in ticket 702:

"Since our reference site will have default message actions configured, but not all clients will be using advocacy, we'd like to hide the 'Message Actions' section of the Advocacy: Actions view from Springboard Admins."

has not been directly addressed. If we don't want Springboard Admins to see the "message actions" section of the advocacy landing page, then a configuration change must be made: springboard admins should not have permission to "create" or "edit any" messages actions, and they must not have "bypass node access" permissions. Either that, or the message actions module must be disabled. 

**Updates to the Springboard Menus and linked pages:**

1. Access to the "Marketing" menu item has been reconfigured. Instead of "access springboard dashboard", access to this area now requires that the user has to ability to manage any of its sub-items. @todo - perhaps this is better managed by a new named permission.

2. The pages under the "Reports" and "Marketing" menu items are aggregates of links to other functions. These links previously displayed to all users, whether they had access to these management functions or not. An access check has been added to each aggregated link.

3. The other pages in Springboard which share this same aggregated link function, such as "springboard/settings/config" are also updated to check access.

4. Access to the "Reports" top-level menu item remains controlled by "Access Springboard Dashboard". @todo - make behave similar to "marketing". The "contacts" view under this item uses the same permission, so changing the top-level perm is moot, for now.


springboard/modules/springboard_admin/springboard_admin.module
```diff
@@ -447,6 +447,130 @@ function springboard_admin_menu_title_callback($original_title, $new_title) {
...
+ * on subsets of other named permissions.
+ *
+ * @return bool
+ */
+function springboard_admin_access_donation_form_pages($type = NULL) {
+  if (user_access("administer springboard")
+    || user_access('administer springboard dashboard')
+    || user_access('administer fundraiser')
+    || user_access('bypass node access')
+    || user_access('edit any donation form')
```
epicflux: @kljr I think this last permission check could be removed as that permission is no longer granted. From here and line 472 below.

PR #1427, opened 2017-04-24
7.x 4.x 1529 fundriaser recurring status

kljr:
Adds a "Status" field to the fundraiser recurring series entity which can be mapped and synced to custom field in .

New series get a status of 'active'. When the final active donation in a series is updated, the series status is updated to reflect the appropriate status based on the $donation->status value of the last recurring donation.

### DIFF
--- a/fundraiser/modules/fundraiser_sustainers/fundraiser_sustainers.module
+++ b/fundraiser/modules/fundraiser_sustainers/fundraiser_sustainers.module
@@ -3909,6 +3909,15 @@ function fundraiser_sustainers_update_recurring_series_entity($master_did) {
       }
     }
 
+    // If the status is canceled, check if its auto_canceled.
+    if (
+      !empty($status)
+      && $status == 'canceled'
+      && $last_donation->recurring->gateway_resp == 'auto_canceled'
+    ) {
+      $status = 'auto_canceled';
+    }
+
     if (!empty($status)) {
       $fundraiser_sustainers_series->status = $status;
       $fundraiser_sustainers_series->save();
--- a/fundraiser/modules/fundraiser_sustainers/fundraiser_sustainers.module
+++ b/fundraiser/modules/fundraiser_sustainers/fundraiser_sustainers.module
@@ -3296,6 +3296,7 @@ function fundraiser_sustainers_fundraiser_donation_success($donation) {
           'did' => $donation->did,
           'uid' => $donation->uid,
           'amount' => $donation->donation['amount'],
+          'status' => 'active',
           'installments' => _fundraiser_sustainers_calculate_installments($donation),
         );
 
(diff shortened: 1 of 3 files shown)


fundraiser/modules/fundraiser_sustainers/fundraiser_sustainers.module
```diff
@@ -3862,12 +3863,63 @@ function _fundraiser_sustainers_update_recurring($donation) {
...
+      static $mdid;
+
+      // Check if there are remaining recurring donations. If there are,
+      // we only want to update once, even if we're cycling through a cancelled
+      // series (if cancelled, we'll pick it up again later).
+      $recur_remain = _fundraiser_sustainers_get_donations_recurr_remaining($record['master_did']);
+      if (!empty($recur_remain)) {
+        if ($record['master_did'] != $mdid && empty($record['cancellation_reason'])) {
+          $fundraiser_sustainers_series = entity_load_single('fundraiser_sustainers_series', $record['master_did']);
+          // Do we need to do this at all?
```
epicflux: @kljr If we go with the update we don't need to do this. An active series that has remaining charges would not need to get updated and saved.

fundraiser/modules/fundraiser_sustainers/fundraiser_sustainers.module
```diff
@@ -3862,12 +3863,63 @@ function _fundraiser_sustainers_update_recurring($donation) {
...
+
+      // Save the master did to a static var in case we're cycling through
+      // a set of donations (normal processing and cancellation of series),
+      // so that we don't repeatedly re-save the sustainers_series object.
+      static $mdid;
+
+      // Check if there are remaining recurring donations. If there are,
+      // we only want to update once, even if we're cycling through a cancelled
+      // series (if cancelled, we'll pick it up again later).
+      $recur_remain = _fundraiser_sustainers_get_donations_recurr_remaining($record['master_did']);
```
epicflux: You could load the series entity at this point as well; since it gets loaded later with either condition. You could then use its status to determine whether or not to operate on it.

fundraiser/modules/fundraiser_sustainers/fundraiser_sustainers.module
```diff
@@ -3862,12 +3863,63 @@ function _fundraiser_sustainers_update_recurring($donation) {
...
+          if ($fundraiser_sustainers_series->status != 'active') {
+            $fundraiser_sustainers_series->status = 'active';
+            $fundraiser_sustainers_series->save();
+            $mdid = $record['master_did'];
+          }
+        }
+      }
+      else {
+        // We only get here if this is the last donation in a series.
+        $statuses = array(
```
epicflux: Having these status values returned by a function would allow us to easily use them in other functions.

fundraiser/modules/fundraiser_sustainers/fundraiser_sustainers.module
```diff
@@ -3862,12 +3863,63 @@ function _fundraiser_sustainers_update_recurring($donation) {
...
   }
   else {
+
     // Merge data together so we get everything in the record.
     $record = array_merge((array) $original, $donation);
     drupal_write_record('fundraiser_sustainers', $record, 'did');
-
     $log = fundraiser_sustainers_log();
     $log->log($donation, $original);
+
```
epicflux: _fundraiser_sustainers_update_recurring() is the right function to update the series entity, but if we move the code related to this operation into its own function that would help to isolate it for testing and use in other places.

PR #1428, opened 2017-04-24
7.x 4.15.1 beta 1808 add log entry for all webform submissions

jcestep:


pcave:
Does the timing of this work correctly? Would springboard_webform_submission_insert run before the submit handler?

scottatdrake:
@pcave What does the update impact tag on this ticket indicate?

scottatdrake:
@pcave oh, it probably means add "Enable the new Springboard Webform" module to the deployment instructions. Will do.

PR #1433, opened 2017-04-27
1826 Recurring minimum donation amount validation update

jesswalls:
Looks like with the dual-ask changes when a form is set to recurring-only, the other field is spoofed to match the form keys for what would be present on a one-time form. Consequently, if there is a recurring_minimum_donation_amount configured on a recurring-only form, the validation for the other amount uses what is configured for the one-time minimum.

This update adds some logic to check for wether the current form is recurring-only to ensure the correct minimum donation amount is used for validation on recurring-only forms.


fundraiser/modules/fundraiser_sustainers/fundraiser_sustainers.module
```diff
@@ -1520,9 +1520,11 @@ function _fundraiser_sustainers_other_amount_field_validate($form, $form_state,
 function _fundraiser_sustainers_recurring_other_amount_field_validate($form, $form_state, $submission_fields, $value) {
   // This validation is only relevant if making a recurring donation via the recurring-amount fields.
   if (
-    empty($form['#node']->recurring_dual_ask_amounts)
+    $form['#node']->recurring_setting !== 'always'
+    && 
+    (empty($form['#node']->recurring_dual_ask_amounts)
     || (is_array($submission_fields['recurs_monthly']) ? empty($submission_fields['recurs_monthly']['recurs'])
```
epicflux: @jesswalls while you did not originally implement this single line if statement inside of a multi line if statement, could you please add a comment so other developers can understand what is being compared.

jesswalls: Commented added per your suggestion.

PR #1436, opened 2017-04-27
7.x 4.x braintree paypal compact

jesswalls:
Adds a usability change for the Braintree Paypal payment method, configurable on a per form basis. When the 'compact' option is enabled, the standard default fundraiser billing and user profile fieldsets are hidden and populated with data returned from Paypal through Braintree; this reduces the amount of information a potential donor needs to fill out before submitting a donation form.


braintree_integration/modules/braintree_payment/js/braintree_payment.paypal.admin.js
```diff
@@ -0,0 +1,47 @@
+(function ($) {
+  Drupal.behaviors.brainTreePayPalCompact = {
+    attach: function (context, settings) {
+      var errors = Drupal.settings.paypalCompactErrors || [];
+      //if (typeof(Drupal.settings.paypalCompactErrors) != 'undefined') {
```
scottatdrake: Hi, @jesswalls. Did you intend to commit this commented-out conditional? If so, it would probably be better to delete it.

epicflux:
This PR is now being handled in #1403.

PR #1442, opened 2017-05-03
1828 Adds error handling for message tracking endpoint.

scottatdrake:


CammoKing:
@scottatdrake This is looking good to me.

PR #1451, opened 2017-05-09
1866 dw skipped records

scottatdrake:
Some records don't have created timestamps, so sorting queries by that field is problematic. This can cause records to get skipped when queuing them up for the DW. 

Sort by  ID instead.


CammoKing:
This looks good Scott.

PR #1463, opened 2017-06-07
1890: Fix compact flow checkbox inclusion on forms created before bra…

robertromore:
…intree applepay+paypal methods enabled


braintree_integration/modules/braintree_payment/inc/paypal.braintree_payment.inc
```diff
@@ -6,6 +6,7 @@
  */
 
 define('BRAINTREE_PAYMENT_PAYPAL_PAYMENT_METHOD_ID', 'braintree_paypal');
+define('BRAINTREE_PAYMENT_PAYPAL_PAYMENT_RULE_NAME', 'braintree_paypal|commerce_payment_braintree_paypal');
```
epicflux: This constant should not be used in the code as we can't safely rely on this rule name always being the same. Payment gateways can be created and cloned causing different rule names. The function _braintree_is_paypal_payment_method() does an appropriate check on the method_id portion of the rule name, which does stay the same.

braintree_integration/modules/braintree_payment/inc/paypal.braintree_payment.inc
```diff
@@ -246,6 +251,19 @@ function _braintree_payment_paypal_method_is_option($node, $form = NULL) {
   if (isset($node->gateways) && isset($node->gateways['paypal']) && _braintree_is_paypal_payment_method($node->gateways['paypal']['id'])) {
     return TRUE;
   }
```
epicflux: This function starts by looking for the paypal option being present on the node, but then switches to looking for the applepay method on the node and/or form. Perhaps we need to create another function for applepay, or rename and move this function and bring the two gateways inline.

braintree_integration/modules/braintree_applepay/braintree_applepay.module
```diff
@@ -230,8 +230,12 @@ function braintree_applepay_form_webform_client_form_alter(&$form, &$form_state)
  * Helper function to determine if method ID is apple pay.
  */
 function _braintree_is_applepay_payment_method($method) {
+  if (empty($method)) {
+    return FALSE;
+  }
+
   list($method_id, $rule_name) = explode('|', $method);
-  return $method_id == BRAINTREE_APPLEPAY_PAYMENT_METHOD_ID;
+  return $method_id == BRAINTREE_APPLEPAY_PAYMENT_METHOD_ID || empty($method_id);
```
epicflux: Do we need the additional check for empty()? This would return TRUE if the method_id was empty, which would seem incorrect.

braintree_integration/modules/braintree_payment/inc/paypal.braintree_payment.inc
robertromore: I've adjusted this in both paypal and apple pay to remove the rule names and instead check if the payment method name (braintree_paypal or braintree_applepay) is part of a key in the array.

PR #1468, opened 2017-06-09
1419: Revising billing update hooks in fundraiser_commerce_cardonfile…

robertromore:
… to support refactoring of billing update form function in fundraiser.module


fundraiser/modules/fundraiser_commerce_cardonfile/fundraiser_commerce_cardonfile.module
```diff
@@ -256,35 +256,64 @@ function fundraiser_commerce_cardonfile_form_fundraiser_update_billing_form_alte
...
+  $donation = $form['#donation'];
+  $submission_fields = $form_state['values'];
+  $donation->donation = array_merge($donation->donation, $submission_fields);
+  $donation->update_billing = TRUE;
+  $donation->update_payment = TRUE;
+  $donation->update_other_orders = $form['#fundraiser_data']['update other orders'];
+  $form_state['values']['cardonfile']['card'] = $form_state['card'];
+  // Using `clone` to ensure this isn't modified by reference later.
+  $form_state['values']['cardonfile']['old_card'] = clone $form_state['card'];
+  fundraiser_donation_update($donation);
```
epicflux: This submit handler would allow for updating a donation that isn't a future sustainer charge. We don't update billing or address information on settled donations. We'll want to verify earlier in the process, probably in fundraiser_commerce_cardonfile_form_commerce_cardonfile_card_form() whether the card is attached to a recurring donation series or not. If it isn't connected to a sustainer series we may need to consider using a different form build and submit process, one that only targets the card on file record.

PR #1497, opened 2017-07-11
re #1984, #1346 - ensure dual ask is set before using recurring minimum

matt-thomas:
Sustainers doesn't check that a donation form has dual ask enabled before using the recurring minimum on the billing update form. This PR ensures such a check is made. It also sets the fallback variable for validation to variable_get('fundraiser_default_minimum', 10.00), rather than 0.


fundraiser/modules/fundraiser_sustainers/fundraiser_sustainers.module
```diff
@@ -2574,11 +2574,18 @@ function fundraiser_sustainers_donation_amount_form($form, &$form_state, $donati
   );
   $form['min_amount'] = array(
     '#type' => 'hidden',
-    '#value' => isset($donation->node->recurring_minimum_donation_amount) ? $donation->node->recurring_minimum_donation_amount : (isset($donation->node->minimum_donation_amount) ? $donation->node->minimum_donation_amount : 0),
+    '#value' => isset($donation->node->minimum_donation_amount) ? $donation->node->minimum_donation_amount : variable_get('fundraiser_default_minimum', 10.00),
   );
 
   $node_currency = fundraiser_get_currency_from_node($donation->node);
-  $min_amount = isset($donation->node->recurring_minimum_donation_amount) ? $donation->node->recurring_minimum_donation_amount : (isset($donation->node->minimum_donation_amount) ? $donation->node->minimum_donation_amount : variable_get('fundraiser_default_minimum', 10.00));
+  $min_amount = isset($donation->node->minimum_donation_amount) ? $donation->node->minimum_donation_amount : variable_get('fundraiser_default_minimum', 10.00);
```
epicflux: @matt-thomas since its the same code to get the $min_amount here as above, could you move this up above the $form['min_amount'] line and then use this value to also set that fields '#value'.

